PR #1, opened 2020-08-06
Add CMake support

gracicot:
This patch enable CMake support for building the library and installing it. By default, it will build the `examples` and `tests` target just like the current makefile. It also add a `test` target that run tests, and also an `install` target so the library is installed into the system (or any specified directory).

To make building the library easier, I added a `conanfile.txt` so dependencies can be installed automatically. I made a setup that is doing so without affecting how the targets are exported. It can be disabled by `-DAEDIS_USE_CONAN=OFF`. To use it, simply create the build directory (for example, `build/`), cd into it and type `conan install ..` then run CMake `cmake ..`

It also supports using the library from the build tree directly. So a CMake project can add `path/to/aedis/build` to the `CMAKE_PREFIX_PATH` variable and `find_package(aedis)` will simply work without even installing the library. 

Here's a couple things to consider:

 - No version file is exported, so users won't be able to use `find_package(aedis 0.1.1)` syntax to check for version compatibility, since this library is not versioned.
 - The old makefile is still there. It could be removed.
 - The header is not in a `include/aedis` directory, making users `#include <aedis.h>` instead of `#include <aedis/aedis.h>`
   - This can be done without changing the project structure.
 - Conan is enabled by default in the CMake file.


mzimbres:
Will have a look soon, thanks.

mzimbres:
I agree with your points about the versions file, the directory `include/aedis` etc. Those are planned features, you can open issues for that if you like.

PR #35, opened 2022-10-30
Added boost.container.pmr to request.

klemens-morgenstern:


CMakeLists.txt
```diff
@@ -98,8 +101,26 @@ target_compile_features(test_low_level PUBLIC cxx_std_17)
 target_compile_features(test_conn_cancel_run PUBLIC cxx_std_20)
 target_compile_features(test_conn_cancel_exec PUBLIC cxx_std_20)
 
-target_link_libraries(intro_tls OpenSSL::Crypto OpenSSL::SSL)
-target_link_libraries(test_conn_tls OpenSSL::Crypto OpenSSL::SSL)
```
mzimbres, line 102: Why did you remove openssl dependency?

include/aedis/resp3/request.hpp
```diff
@@ -205,11 +207,18 @@ class request {
     *  
     *  @param cfg Configuration options.
     */
-   explicit request(config cfg = config{false, true, false, true})
-   : cfg_{cfg}
+   explicit request(boost::container::pmr::memory_resource * resource)
+         : payload_(resource), cfg_{false, true, false, true}
```
mzimbres, line 211: AFAICS, we don't need this overload, the one below is enough. I wouldn't like to have `config{false, true, false, true}` twice in the code.

mzimbres:
` boost::container::pmr::string` was indeed a good idea.

mzimbres:
Does the serialization example still work? See https://github.com/mzimbres/aedis/blob/ec6e99d99a9cbc62b06a7362904bd343b37ca008/examples/serialization.cpp#L60

mzimbres:
I see two downsides about using `boost::container::pmr::string` instead of `std::string` with template allocator in the request
* All request member functions are template anyway, which means I won't be able to move some code to `cpp` files.
* boost::container is a new dependency that must be built  and linked against. CMake can't find it in my installation and I wouldn't like to force this on users.

I will sketch a solution with std::string based on this PR, let me see how it compares.

CMakeLists.txt
klemens-morgenstern: That would've been by accident.

klemens-morgenstern:
If you're planning to stay on C++17 you can just use std::pmr instead: https://en.cppreference.com/w/cpp/header/memory_resource

mzimbres:
Merged this branch locally after moving to std::pmr::string and other unrelated changes: https://github.com/mzimbres/aedis/commit/f93f3cab58ddec2b90ec353a6b65892135ca94d2

PR #98, opened 2023-05-13
94 unify redisconnection and redissslconnection

mzimbres:


ysc3839:
Can you provide a compile-time option to disable SSL completely?

mzimbres:
@ysc3839 Can you please open a new issue so we can track this more easily? Thanks.

Issue #115, opened 2023-06-20
build config enforces building tests even as third-party


cthulhu-irl:
Currently in cmake file, building tests is enforced, and unlike many other common libraries tests are not guarded by a cmake option variable.

This makes it hard to use as a third-party library, specially as it's quite slow.

mzimbres:
Hi, I will have a look on how to disable building the tests and the examples. Notice however that to use Boost.Redis you have to add the include path and `#include <boost/redis/src.hpp>` in one of your cpp files.

Regarding compilations times: I am about to merge a branch that reduces it by close to 70%.

cthulhu-irl:
I can simply add something similar to the following (from libfmt).

options for particular targets/features/etc.
```cmake
# Options that control generation of various targets.
option(FMT_DOC "Generate the doc target." ${FMT_MASTER_PROJECT})
option(FMT_INSTALL "Generate the install target." ON)
option(FMT_TEST "Generate the test target." ${FMT_MASTER_PROJECT})
```

putting the related section under the condition.
```cmake
if (FMT_TEST)
  enable_testing()
  add_subdirectory(test)
endif ()
```

and have the default build options enabled depending on if it's being built directly or has been included via add_subdirectory as a third-party:
```cmake
# Determine if fmt is built as a subproject (using add_subdirectory)
# or if it is the master project.
if (NOT DEFINED FMT_MASTER_PROJECT)
  set(FMT_MASTER_PROJECT OFF)
  if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
    set(FMT_MASTER_PROJECT ON)
    message(STATUS "CMake version: ${CMAKE_VERSION}")
  endif ()
endif ()
```

looking at the root cmake of this project, the following sections seem to need an option:
 - tests
 - coverage
 - examples
 - docs
 - benchmarks
 - install (including packing)

I can make a PR if the above approach is okay.

mzimbres:
A PR would be welcome. I think there needs to be no option for coverage as that is better handled in the preset. Have a look at https://github.com/boostorg/redis/blob/607946f00e62f4391bc67be6b3e96ebbe63dfc14/CMakePresets.json#L25.

mzimbres:
Merged, thanks for your contribution.

PR #117, opened 2023-06-23
build: add cmake options

cthulhu-irl:
Added cmake options for install, tests, examples, and doc instead of always building them.

options are enabled by default when building the project directly, otherwise if add_subdirectory (directly or by FetchContent, etc), then unnecessary options will be disabled.

This PR is related to #115 issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,15 @@ cmake_minimum_required(VERSION 3.14)
 
 #set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CMAKE_COMMAND} -E time")
 
+# determine whether it's main/root project
+# or being built under another project.
+if (NOT DEFINED BOOST_REDIS_MAIN_PROJECT)
+   set(BOOST_REDIS_MAIN_PROJECT OFF)
+   if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+      set(BOOST_REDIS_MAIN_PROJECT ON)
+   endif()
+endif()
+
 project(
     boost_redis
     VERSION 1.4.1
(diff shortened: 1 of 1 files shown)


mzimbres:
Great, thanks.

mzimbres:
Can you fix the windows build? It looks like the cpp20 code is failing to build.

mzimbres:
I have just merged the branch that improves compilation times, You might want to rebase so that CI and local test will run faster.

cthulhu-irl:
@mzimbres the workflow wasn't kicked to run after rebase, would you run it?

mzimbres:
@cthulhu-irl: Windows builds are still broken.

cthulhu-irl:
> @cthulhu-irl: Windows builds are still broken.

it should be fixed now.

mzimbres:
Humm, not all examples should be run as tests.

cthulhu-irl:
> Humm, not all examples should be run as tests.

would you review the code to see if there are things that don't match the previous version's behavior?

mzimbres:
Will do in the coming week. Examples like cpp20_streams, cpp20_subscriber run forever and should not be run as tests.

CMakeLists.txt
```diff
@@ -2,6 +2,17 @@ cmake_minimum_required(VERSION 3.14)
 
 #set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CMAKE_COMMAND} -E time")
 
+# determine whether it's main/root project
+# or being built under another project.
+if (NOT DEFINED BOOST_REDIS_MAIN_PROJECT)
```
mzimbres, line 7: This check will be always evaluated to false? Why can't we simply write

```
   if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
      set(BOOST_REDIS_MAIN_PROJECT ON)
   else()
      set(BOOST_REDIS_MAIN_PROJECT OFF)
```

mzimbres:
Also, cpp20 executables are being compiled as cpp17. For example, I get
```
$ ./build/g++-11/cpp20_intro
Requires coroutine support.
```
Instead of the actual output.

CMakeLists.txt
hexorer: In case for some reason user needed to define it manually, it would skip reassigning. specially when not using `add_subdirectory` but requiring the same behavior as if it was a subproject, for example using the library from another build system.

I personally haven't came across such situations, and I've just copied and edited it for this project. but my guess is the above.

cthulhu-irl:
> Also, cpp20 executables are being compiled as cpp17. For example, I get
> 
> ```
> $ ./build/g++-11/cpp20_intro
> Requires coroutine support.
> ```
> 
> Instead of the actual output.

@mzimbres I think it should be fixed now. kinda weird that CI didn't yield the same error while running tests.

```
cthulhu@lolnet:~/projects/boost-redis$ ./build/g++-11/cpp20_intro
(Boost.Redis) Resolve results: 127.0.0.1:6379
(Boost.Redis) Connected to endpoint: 127.0.0.1:6379
(Boost.Redis) Bytes written: 117
(Boost.Redis) Hello: Success
PING: Hello world
(Boost.Redis) Connection lost: Operation canceled
``

mzimbres:
Looks good. I will have a final look tomorrow and then merge it. Thanks.

PR #174, opened 2023-12-31
Provides a way of passing a custom ssl context to the connection.

mzimbres:
@anarthal It would be great if you had a look at this PR. Thanks.


anarthal:
It looks like moving a `connection` constructed from a null context leaves a dangling pointer to the fallback context. Plus you're creating a context per object even if you don't need one. If you're gonna do like that, better pass the context by value.

I solved both of these issues by having the fallback context as a singleton. Klemens solved it using an Asio service.

mzimbres:
@mzimbres I am glad I pulled you in the PR. I have been hurrying with some implementations as my free time has dropped.

anarthal:
I also recommend using `ssl::context::tlsv12_client` as the default method. This disables support for TLS 1.0 and 1.1, which are considered vulnerable.

PR #201, opened 2024-06-09
Adds a type-erased response adapter to the public API

### LINKED ISSUE #128: Move the adapter module to the detail namespace
No elements of the adapter module appears in the public API so it should be moved in the detail namespace and directories. Some parts of it belong into `redis/detail` while other in `redis/resp3/detail`.

anarthal:
Would something along these lines suit you @mzimbres? If it does, I'll clean it up and write docs & tests for it.

close #128


include/boost/redis/detail/connection_base.hpp
```diff
@@ -440,21 +442,26 @@ class connection_base {
...
          < CompletionToken
          , void(system::error_code, std::size_t)
          >(exec_op<this_type>{this, info}, token, writer_timer_);
    }
 
+   template <class Response, class CompletionToken>
+   auto async_exec(request const& req, Response& resp, CompletionToken token)
+   {
+      return async_exec(req, any_adapter(resp), std::forward<CompletionToken>(token));
+   }
```
mzimbres: This overload is not needed, the `connection_base` class is not meant to be used directly.

include/boost/redis/connection.hpp
```diff
@@ -394,7 +406,13 @@ class connection {
    template <class Response, class CompletionToken>
    auto async_exec(request const& req, Response& resp, CompletionToken token)
    {
-      return impl_.async_exec(req, resp, std::move(token));
+      return async_exec(req, any_adapter(resp), std::move(token));
+   }
+
+   template <class CompletionToken>
+   auto async_exec(request const& req, any_adapter adapter, CompletionToken token)
```
mzimbres: This function should now be implemented in terms of any_completion_handler so that the instantiation can be moved to the .ipp file, similar to how async_run is implemented (see above). That is IMO the main motivation for these changes: reduce compilation times.

mzimbres:
@anarthal Thanks, see my comments.

anarthal:
Did you publish your comments? I can't see any of them

mzimbres:
Are you still going to work on this PR?

anarthal:
whoa, I completely forgot about this! Yes, I'll finish it this incoming week.

anarthal:
@mzimbres this is now complete on my side. Please let me know if any changes are required.

mzimbres:
Many thanks! Give me some days to review it. I would also say you this is worth listing in the Changelog.

include/boost/redis/connection.hpp
```diff
@@ -392,9 +408,21 @@ class connection {
...
-   auto async_exec(request const& req, Response& resp, CompletionToken token)
+   auto async_exec(request const& req, Response& resp, CompletionToken&& token)
    {
-      return impl_.async_exec(req, resp, std::move(token));
+      return async_exec(req, any_adapter(resp), std::forward<CompletionToken>(token));
+   }
+
+   /// Calls `boost::redis::basic_connection::async_exec`.
+   template <class CompletionToken>
+   auto async_exec(request const& req, any_adapter adapter, CompletionToken&& token)
```
mzimbres, line 418: I did not document the adapters and I don't see any immediate need if this overload for users, The main idea of this PR is to improve compilation times. I think we should wait until someone asks for an adapter interface, that will only happen when people want to pass their own data structures in.

Edit: Make adapter overloads private for now.

include/boost/redis/adapter/any_adapter.hpp
```diff
@@ -0,0 +1,80 @@
...
+    static auto create_impl(T& resp) -> impl_t
+    {
+        using namespace boost::redis::adapter;
+        auto adapter = boost_redis_adapt(resp);
+        std::size_t size = adapter.get_supported_response_size();
+        return { std::move(adapter), size };
+    }
+
+    template <class Executor>
+    friend class detail::connection_base;
```
mzimbres, line 61: I think there should be no `friend` class here, the class should define 

```cpp
void operator(std::size_t, resp3::basic_node<std::string_view> const&, system::error_code&)
```

and forward the call directly to `impl_`. Once you do that you can also remove [this](https://github.com/boostorg/redis/blob/e8dd4d69eb3524c9a0efb0c9723790b479a4ba8f/include/boost/redis/detail/connection_base.hpp#L515) typedef and replace it with `any_adapter` AFAICS.

In a later issue we will have to think about how this relates to the [receiver adapter](https://github.com/boostorg/redis/blob/e8dd4d69eb3524c9a0efb0c9723790b479a4ba8f/include/boost/redis/detail/connection_base.hpp#L516) or whether we need another any adapter class. This will be needed to type erase the receive operation too

include/boost/redis/connection.hpp
anarthal: Wasn't this PR actually motivated by a user asking for this exact functionality? See comments in https://github.com/boostorg/redis/issues/128

include/boost/redis/adapter/any_adapter.hpp
anarthal: This makes sense only if `any_adapter` is a public class - otherwise I can remove it (together with the Doxygen docs). See my question above.

mzimbres: Ok, let us keep it public.

anarthal:
This is ready

PR #217, opened 2024-10-19
212 async exec never returns when the network is disconnected and then restored

### LINKED ISSUE #212: async_exec never returns when the network is disconnected and then restored.
In scenarios where the connection is broken and then restored, the redis::connection does not always return to a normal state. I have not delved into the internal implementation, which may be related to the coroutine scheduling mechanism of asio. The following simple example code can reproduce the issue:

```cpp
constexpr auto use_nothrow_awaitable =
    asio::experimental::as_tuple(asio::deferred);

awaitable<void> Test() {
  redis::config cfg;
  cfg.addr.host = "";
  cfg.addr.port = "6379";
  cfg.username = "";
  cfg.password = "";
  cfg.use_ssl = false;

  conn_ = std::make_shared<redis::connection>(io_context.get_executor());
  conn_->async_run(cfg, {}, asio::consign(asio::detached, conn_));

  for (;;) {
    redis::request req;
    req.push("PING", "Hello");
    redis::response<std::string> resp;
    // normal
    co_await conn_->async_exec(req, resp, use_nothrow_awaitable);
    // probability never return
    //co_await conn_->async_exec(req, resp, asio::use_awaitable);
    // probability never return
    //co_await conn_->async_exec(req, resp, asio::deferred);
    LOG_INFO("resp [%s]", std::get<0>(resp).value().c_str());

    asio::steady_timer timer(co_await asio::this_coro::executor);
    timer.expires_after(std::chrono::seconds(1));
    co_await timer.async_wait(use_nothrow_awaitable);
  }
  co_return;
}
```
version:boost1.86

The method to simulate the disconnection and restoration of the network is through firewall commands:
```cpp
iptables -A OUTPUT -p tcp --dport=6379 -j DROP
iptables -D OUTPUT -p tcp --dport=6379 -j DROP
```

Additionally, as mentioned [here](https://github.com/boostorg/redis/issues/199), ```cppusing cancel_if_not_connected = true``` with asio::deferred is also fine, but I believe this may not be a good solution.

mzimbres:


include/boost/redis/detail/resolver.hpp
```diff
@@ -40,7 +40,9 @@ struct resolve_op {
 
          resv_->results_ = res;
 
-         // TODO: map operation_canceled into error::resolve_timeout
+         if (ec == asio::error::operation_aborted) {
+            ec == error::resolve_timeout;
```
bruno-viva, line 44: Should this be `ec = error::resolve_timeout`?

PR #228, opened 2024-12-23
Uses asio::deferred_t as default completion type.

### LINKED ISSUE #225: Support default completion tokens in redis::connection
The non-templated connection should have `asio::deferred` as the default completion token for all its operations, so the following works:

```cpp
redis::connection conn (ctx);
// ...
co_await conn.async_exec(req, res);
```

mzimbres:


mzimbres:
Hi @anarthal, does this fix the issue? Thanks.

anarthal:
From what I read from my cell phone I'd say it does. I'm planning on reviewing it from my PC on Monday?

include/boost/redis/connection.hpp
```diff
@@ -1250,8 +1251,8 @@ class connection {
       { return impl_.get_executor(); }
 
    /// Calls `boost::redis::basic_connection::async_run`.
-   template <class CompletionToken>
-   auto async_run(config const& cfg, logger l, CompletionToken token)
+   template <class CompletionToken = asio::deferred_t>
+   auto async_run(config const& cfg, logger l, CompletionToken token = {})
```
anarthal: As minor comments, consider unifying all signatures using completion tokens to `CompletionToken&&`

include/boost/redis/connection.hpp
```diff
@@ -1281,15 +1274,19 @@ class connection {
    }
 
    /// Calls `boost::redis::basic_connection::async_exec`.
-   template <class Response, class CompletionToken>
-   auto async_exec(request const& req, Response& resp, CompletionToken&& token)
+   template <class Response, class CompletionToken = asio::deferred_t>
```
anarthal, line 1278: You can also consider using `BOOST_ASIO_COMPLETION_TOKEN_FOR` to implement concept checks.

anarthal:
I've left a couple of suggestions. They're not at all needed for this to work, you can do them if/whenever you want.

Also, I'd also write a couple of tests to cover that all the functions you changed successfully support the default token. That might be even a piece of code that you don't run, just to check it builds.

As I said, they're suggestions, the changes look good as they are.

PR #232, opened 2025-01-27
Remove executable flags from key and certificate files

striezel:
The certificate and key files should not have the executable flag set.

See <https://github.com/boostorg/admin/issues/47#issuecomment-2575165830>.


mzimbres:
@striezel Thanks for the PR. I think these files were added in a PR from @anarthal so I am requesting his review as well.

anarthal:
Looks good to me. I merged the equivalent PR in MySQL and everything was good.

PR #239, opened 2025-04-07
Responses now admit all integer types again

### LINKED ISSUE #238: Boost 1.88 breaks conversions to some integer types
Testing with Boost 1.88.0 RC1, Ubuntu 24.04, both gcc-14 and clang-18. The following code used to compile under Boost 1.87:

```cpp
#include <boost/asio/detached.hpp>
#include <boost/asio/io_context.hpp>
#include <boost/redis.hpp>

namespace redis = boost::redis;
namespace asio = boost::asio;

int main()
{
    asio::io_context ctx;
    redis::connection conn(ctx);
    redis::request req;
    redis::response<long> res;
    conn.async_exec(req, res, asio::detached);
}
```

This is no longer the case, since Boost.Redis seems to think that a `long` is not an integral type.

The offending code is:

https://github.com/boostorg/redis/blob/ac4e6e29b372783b9c11b015ef187f07f2112e46/include/boost/redis/adapter/detail/adapters.hpp#L40-L44

Integer types must be 8, at least, and only 3 are listed there. What's the rationale behind defining a custom type trait for integers, rather than `std::is_integral`? (add a `!std::is_same_v<T, bool>` if you need to exclude booleans).

anarthal:
Responses can now be used again with signed char, unsigned char, short, unsigned short, unsigned int, long, and unsigned long.
Added unit tests for conversions to integers, bool and double

close #238


mzimbres:
Thanks @anarthal

@mclow Can I merge my `develop` into the `master` so it contains the fix for the release?

PR #249, opened 2025-05-12
Adds automatic formatting with clang-format

anarthal:


anarthal:
Merging as per @mzimbres request

Issue #140, opened 2023-08-15
detail::exec_op::info memory not freed before completion


anarthal:
Memory for `req_info_type` in `exec_op` is being allocated using the handler's associated allocator, but is not being freed before calling `complete`. This is a violation of one of Asio's composed ops principles.

This has been caught by ubsan when using Boost.Redis together with Boost.Async (I'm exercising the library by rewriting BoostServerTech). Boost.Async uses an internal, per-operation allocator. That is, when you call `conn.async_exec(whatever, use_op)`, a stack-based memory pool of 2kb is created. All allocations for this `async_exec` will use these 2kb of memory first. Once this `async_exec` operation is done, the pool is destroyed. However, the call `self.complete` within `async_exec` resumes and keeps executing the coroutine that called `async_exec`. At this point, `async_exec` has a `req_info_type` object pointing to memory that has been freed (or re-used). When the `shared_ptr<req_info_type>` destructor runs, it will be accessing freed memory.

It's likely that this can be used to build exploits leading to arbitrary code execution.

anarthal:
Even if you use the standard allocator, I'd say you should free any temporary memory acquired by the operation before calling complete, unless there is a strong reason not to do so, so that allocations don't stack up. But probably Klemens can give you a better understanding of the subject.

anarthal:
This is no longer a problem since the allocations of multiplexer::elem use the standard allocator, and are no longer considered transient. This is, under certain circumstances (e.g. we get a cancellation and we must leave the elem allocated after invoking the handler). #250 will make this better, but still, allocations are non-transient at this point, so this issue no longer makes sense.

PR #250, opened 2025-05-12
async_exec now uses a sans-io strategy

anarthal:
async_exec now supports partial and total cancellation when the request is pending
Added docs on async_exec's cancellation support
Added detail::exec_fsm and macros for coroutine emulation
Requests are now removed from the multiplexer when an error is encountered

### DIFF
--- a/include/boost/redis/detail/exec_fsm.hpp
+++ b/include/boost/redis/detail/exec_fsm.hpp
@@ -9,13 +9,9 @@
 #ifndef BOOST_REDIS_EXEC_FSM_HPP
 #define BOOST_REDIS_EXEC_FSM_HPP
 
-#include <boost/redis/detail/coroutine.hpp>
 #include <boost/redis/detail/multiplexer.hpp>
-#include <boost/redis/request.hpp>
 
 #include <boost/asio/cancellation_type.hpp>
-#include <boost/asio/error.hpp>
-#include <boost/assert.hpp>
 #include <boost/system/error_code.hpp>
 
 #include <cstddef>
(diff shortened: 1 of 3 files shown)


include/boost/redis/detail/exec_fsm.hpp
```diff
@@ -0,0 +1,148 @@
...
+         while (true) {
+            // Wait until we get notified. This will return once the request completes,
+            // or upon any kind of cancellation
+            BOOST_REDIS_YIELD(resume_point_, 3, exec_action_type::wait_for_response)
+
+            // If the request has completed (with error or not), we're done
+            if (elem_->is_done()) {
+               return exec_action{
+                  elem_->get_error(),
+                  elem_->get_error() ? 0u : elem_->get_read_size()};
```
mzimbres: I think here you could return whatever is in `elem_->get_read_size()`. Seting to `0` when there is an error seems unnecessary and might indeed erase useful information.

include/boost/redis/detail/exec_fsm.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+namespace boost::redis::detail {
+
+// What should we do next?
+enum class exec_action_type
+{
+   setup_cancellation,  // Set up the cancellation types supported by the composed operation
+   immediate,           // Invoke asio::async_immediate to avoid re-entrancy problems
+   done,                // Call the final handler
+   write,               // Notify the writer task
```
mzimbres: I think the naming here could be more descriptive `notify_writer` for example.

include/boost/redis/impl/multiplexer.ipp
```diff
@@ -119,6 +119,7 @@ std::pair<tribool, std::size_t> multiplexer::consume_next(system::error_code& ec
 
    if (ec) {
       reqs_.front()->notify_error(ec);
+      reqs_.pop_front();
```
mzimbres, line 122: Interesting, this is probably fixing a bug.

anarthal: It probably fixes potential UB in `multiplexer::cancel_on_conn_lost()` when the connection is lost due to a parsing error.

include/boost/redis/impl/exec_fsm.ipp
```diff
@@ -0,0 +1,94 @@
...
+            return exec_action{asio::error::operation_aborted};
+         }
+
+         // If we hit a terminal cancellation, tear down the connection.
+         // Otherwise, go back to waiting.
+         // TODO: we could likely do better here and mark the request as cancelled, removing
+         // the done callback and the adapter. But this requires further exploration
+         if (!!(cancel_state & asio::cancellation_type_t::terminal)) {
+            BOOST_REDIS_YIELD(resume_point_, 5, exec_action_type::cancel_run)
+            elem_.reset();  // Deallocate memory before finalizing
```
mzimbres, line 83: AFAICS this is going to address https://github.com/boostorg/redis/issues/140 so after merge we can close that issue too.

PR #253, opened 2025-05-18
Reconnection no longer causes the writer to perform busy waiting

### LINKED ISSUE #252: Writer degrades to busy polling after reconnection
After a successful reconnection, the writer task revers to doing busy polling, rather than waiting for notifications, yielding in a high CPU usage. This is due to the following lines changing the expiry of the writer timer:

https://github.com/boostorg/redis/blob/328ad97a79cf66108a11294462aa636b5abce927/include/boost/redis/connection.hpp#L445-L447

Timers are stateful, and keep the set expiration time. When a wait is issued against a timer with an expiry date in the past, it completes immediately, hence yielding the aforementioned busy wait situation.

I discovered this while implementing the tests for #250. The fix is trivial - using a separate timer for waits. This also simplifies the logic of #250. I will issue a PR shortly. I've opened the issue to keep track of it, in case anyone has experienced it.

anarthal:
Reconnection now uses a separate timer to wait

close #252

### DIFF
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -442,12 +442,10 @@ public:
             return;
          }
 
-         // It is safe to use the writer timer here because we are not
-         // connected.
-         conn_->writer_timer_.expires_after(conn_->cfg_.reconnect_wait_interval);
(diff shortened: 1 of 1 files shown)


anarthal:
Unfortunately, I don't see a way of adding a regression test for this. I've done a manual test.

PR #254, opened 2025-05-20
Enables -Wall and -Werror in CIs

anarthal:
Fixes some warnings due to unused variables, signed to unsigned comparisons, and superfluous moves


anarthal:
I've split this from #250 to make it more self-contained. -Werror here is useful if using Boost lightweight test (included in Boost.Core), which I think we should use here instead of Boost.Test (it's much more lightweight). In lightweight test, tests are just functions that get called from main(). Building with -Werror makes uncalled tests an error.

In any case, it's useful that header-only libraries ship with as few warnings as possible, since they get propagated to user code.

anarthal:
Merging as requested by Marcelo.

Issue #246, opened 2025-05-02
unix sockets support?


eaziz4:
I don't see how to connect to my redis instance via unix socket.  I only see the host + port option in the config class.  Is this supported?

mzimbres:
Hi @eaziz4, this is not yet supported. I would like to implement this feature but don't have much time available at the moment, so I can't make commitments about when I will do it.

eaziz4:
No worries.  Can this issue stay open until it is implemented as I would very much prefer to use this library over others, but need socket support as performance is very important for me.

anarthal:
@mzimbres both `connection` and `basic_connection<Executor>` have a `next_layer()` member that returns the underlying `asio::ssl::stream<asio::ip::tcp::socket>`. To support UNIX sockets, we need to either remove this member (backwards incompatible), change it to `asio::ssl::stream<asio::generic::socket>` (backwards incompatible) or make a new `basic_connection_v2` and `connection_v2` (maintenance problems). What are your views on this?

PR #256, opened 2025-05-21
Marks the next_layer() functions as deprecated

anarthal:


anarthal:
This prepares the library to implement #246 (ETA Boost 1.90).
Note that `BOOST_DEPRECATED` is used instead of `[[deprecated/(M)]]` because it supports being disabled by defining `BOOST_ALLOW_DEPRECATED`.

PR #259, opened 2025-05-25
Increases the robustness of integration tests

anarthal:
Updates multiplexer to make requests complete with `error_code()` rather than `error_code(0)`
Integration tests now use `io_context::run_for` to run with a timeout and avoid deadlocks
Tests now use concrete lambdas where generic ones are not required
Tests now use BOOST_TEST with operator== to print values on error
Tests now use anonymous namespaces to detect dead code
Adds run_coroutine_test and removed start from common.hpp
Updates test_reconnect to perform relevant checks
Refactors how test_issue_50 launches its coroutines
Groups tests in CMake as unit or integration
Updates Jamfile tests to contain all unit tests and only unit tests


test/test_conn_echo_stress.cpp
```diff
@@ -43,90 +49,106 @@ std::ostream& operator<<(std::ostream& os, usage const& u)
...
+   conn.cancel();
 }
 
-auto echo_session(std::shared_ptr<connection> conn, std::shared_ptr<request> pubs, int n)
-   -> net::awaitable<void>
+auto echo_session(connection& conn, const request& pubs, int n) -> net::awaitable<void>
 {
    for (auto i = 0; i < n; ++i)
-      co_await conn->async_exec(*pubs, ignore, net::deferred);
+      co_await conn.async_exec(pubs, ignore);
```
mzimbres, line 82: `ignore` is the [default response type](https://github.com/boostorg/redis/blob/0c8c6fcc095ff103e9adda708c1efb06fabfe08d/include/boost/redis/connection.hpp#L665) so this could be written more simply as `co_await conn.async_exec(pubs);`

I have just noted that although `ignore` is the default response type in `basic_connection` it isn't in [connection](https://github.com/boostorg/redis/blob/0c8c6fcc095ff103e9adda708c1efb06fabfe08d/include/boost/redis/connection.hpp#L901). I will open an issue to fix this.

anarthal: Addressed in #265

PR #267, opened 2025-06-07
Restores the TLS tests in CI

anarthal:
Renews test certificates
Recovers and refactors test_conn_tls.cpp
Adds a test for TLS reconnection


anarthal:
This PR entails no functional change. It makes TLS in the CI Redis servers functional, restores the TLS tests that we had, and adds some extra checks. This is going to be handy to test the changes I need to do to implement UNIX sockets properly (they involve refactoring this part).

mzimbres (approved):

PR #272, opened 2025-06-15
Adds support for UNIX sockets

### LINKED ISSUE #246: unix sockets support?

anarthal:
close #246


example/cpp20_unix_sockets.cpp
```diff
@@ -0,0 +1,60 @@
...
+using boost::redis::logger;
+using boost::redis::connection;
+
+#ifdef BOOST_ASIO_HAS_LOCAL_SOCKETS
+
+auto co_main(config cfg) -> asio::awaitable<void>
+{
+   // If unix_socket is set to a non-empty string, UNIX sockets will be used
+   // instead of TCP. Set this value to the path where your server is listening.
+   // UNIX connections work in the same way as TCP connections.
```
mzimbres: _UNIX connections_ should be _UDS connections_

example/cpp20_unix_sockets.cpp
```diff
@@ -0,0 +1,60 @@
...
+   conn->cancel();
+
+   std::cout << "Response: " << std::get<0>(resp).value() << std::endl;
+}
+
+#else
+
+auto co_main(config) -> asio::awaitable<void>
+{
+   std::cout << "Sorry, your system does not support UNIX sockets\n";
```
mzimbres: UNIX Domain Sockets instead of UNIX sockets

include/boost/redis/detail/redis_stream.hpp
```diff
@@ -17,31 +17,48 @@
...
+
+   transport_type transport_{transport_type::tcp};
+   bool ssl_stream_used_{false};
 
    void reset_stream() { stream_ = {resolv_.get_executor(), ssl_ctx_}; }
 
+   static transport_type transport_from_config(const config& cfg)
+   {
+      return cfg.unix_socket.empty() ? (cfg.use_ssl ? transport_type::tcp_tls : transport_type::tcp)
+                                     : transport_type::unix_socket;
```
mzimbres: I think this would be more readable as an if-else chain than ternary operators.

include/boost/redis/detail/redis_stream.hpp
```diff
@@ -139,7 +176,10 @@ class redis_stream {
    : ssl_ctx_{std::move(ssl_ctx)}
    , resolv_{ex}
    , timer_{ex}
-   , stream_{std::move(ex), ssl_ctx_}
+   , stream_{ex, ssl_ctx_}
+#ifdef BOOST_ASIO_HAS_LOCAL_SOCKETS
+   , unix_socket_{std::move(ex)}
```
mzimbres: I don't see the point of moving the executor here instead of copying since it is a lightweight object. The stream_ parameter above is copying it too.

anarthal: This is what the theory says, but unfortunately `asio::any_io_executor` does make a difference between copying and moving (think of it as a `std::function`): see how its base class [move](https://github.com/boostorg/asio/blob/62864c570afcb7234442168ecacd53af701d7d9e/include/boost/asio/execution/any_executor.hpp#L614-L631) and [copy](https://github.com/boostorg/asio/blob/62864c570afcb7234442168ecacd53af701d7d9e/include/boost/asio/execution/any_executor.hpp#L559-L573) differ. The performance difference is probably insignificant though.

The stream copies it because it can only be moved once - after the move, the object becomes invalid (as a `std::function` would).

example/cpp20_unix_sockets.cpp
anarthal: Changed wording

include/boost/redis/detail/redis_stream.hpp
anarthal: I've updated this to make the move always happen, even if no UNIX sockets are supported.

PR #273, opened 2025-06-15
Updates the Logger interface to allow extensibility and type erasure

### LINKED ISSUE #213: Allow integrating third party logger
For pretty much any production usage you'd want to redirect all logging into the application log file (not std::cerr).
Would be great to have a way to integrate a third party logger, can be as simple as a <severity, message> callback.

anarthal:
Removes all the logger::on_xxx functions
Removes the Logger template parameter to async_run
Adds a logger constructor that allows passing a std::function to customize logging behavior
Adds constructors to connection and basic_connection taking a logger
Deprecates config::logger_prefix
Deprecates the async_run overload taking a logger parameter
Deprecates the basic_connection::async_run overload not taking any config object
Deprecates the basic_connection::next_layer_type typedef
Makes the default log level logger::info
Makes the logging thread-safe
Cleans up deprecated functionality from examples
Adds docs on logging
Adds an example on how to integrate spdlog into Boost.Redis logging

close #213

### DIFF
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -579,7 +579,7 @@ public:
       "and use the other async_run overloads.")
    auto async_run(config const& cfg, logger l, CompletionToken&& token = {})
    {
-      set_run_logger(l, cfg);
+      set_stderr_logger(l.lvl, cfg);
       return async_run(cfg, std::forward<CompletionToken>(token));
    }
 
@@ -826,9 +826,9 @@ private:
    }
 
    // Used by both this class and connection
-   void set_run_logger(const logger& lgr, const config& cfg)
+   void set_stderr_logger(logger::level lvl, const config& cfg)
    {
-      logger_.reset(detail::make_stderr_logger(lgr.lvl, cfg.log_prefix));
+      logger_.reset(detail::make_stderr_logger(lvl, cfg.log_prefix));
    }
 
    template <class> friend struct detail::reader_op;
--- a/include/boost/redis/impl/connection.ipp
+++ b/include/boost/redis/impl/connection.ipp
@@ -32,7 +32,7 @@ void connection::async_run_impl(
 {
    // Avoid calling the basic_connection::async_run overload taking a logger
    // because it generates deprecated messages when building this file
-   impl_.set_run_logger(l, cfg);
+   impl_.set_stderr_logger(l.lvl, cfg);
    impl_.async_run(cfg, std::move(token));
 }
 


include/boost/redis/connection.hpp
```diff
@@ -516,18 +512,19 @@ class basic_connection {
     *  For example on how to call this function refer to
     *  cpp20_intro.cpp or any other example.
     */
-   template <
-      class Logger = logger,
-      class CompletionToken = asio::default_completion_token_t<executor_type>>
-   auto async_run(config const& cfg = {}, Logger l = Logger{}, CompletionToken&& token = {})
+   template <class CompletionToken = asio::default_completion_token_t<executor_type>>
+   auto async_run(config const& cfg = {}, logger l = {}, CompletionToken&& token = {})
```
mzimbres: I think that if the logger becomes a class member we should remove it form `async_run` and set it elsewhere i.e. constructor or `set_logger`, otherwise I am afraid `async_exec` and `async_run` might race e.g. user first call `async_exec` when no logger is set.

mzimbres:
I think the current approach used for logging does not scale well since we have to add a new function everywhere logging is required.  That means, everytime the code is refactored or a new features such as sentinel and cluster support is added, the logger interface will have to be changed or externed. Sometimes it might be necessary to log types from the `detail` namespace which cannot appear in the logging interface.

Considering this I would like to make the log interface stupid, for example

```cpp
struct logger {
   // Syslog levels.
   enum class level {disabled, emerg, alert, crit, err, warning, notice, info, debug};

   // Log object passed by the user.
   std::function<void(level, std::string_view)> log;

   // User decides up to which level he wants to see messages.
   level level_;
};
```

and use it like

```cpp
logger l;

l.log = [](level l, std::string_view msg)
{
   if (l >= level_)
      std::clog << msg << std::endl;
};

l.level_ = logger::level::crit;
```

Once the logger is this stupid Boost.Redis would then only have to care about message formatting and we can do that on our side with macros to avoid parameter evaluation, for example

```cpp
#define BOOST_REDIS_LOG_ERROR_CODE(logger, level, ec)\
   if (level >= logger_.get_max_level() && ec) {\
      std::string tmp;\
      tmp += ec.message();\
      tmp += " [";\
      tmp += ec.to_string();\
      tmp += ']';\
      logger.log(tmp)\
   }\
```

What do you think? Can we take that direction in this PR?

anarthal:
This is absolutely the direction it is already taking. The logging functions are actually in the detail namespace, so they're no longer part of the interface. My plan is gradually removing them as we migrate things to the sans-io approach. Each fsm would then be responsible for formatting. I've just left them with a similar interface as we used to have to reduce the amount of changes, and because fsms don't exist yet for the algorithms that log.

include/boost/redis/connection.hpp
anarthal: We currently don't log anything in async_exec. If you're OK with it, I will add it as a constructor argument, and deprecate passing it to `async_run`. And we won't log anything in async_exec until we remove the deprecated function.

anarthal:
This is now complete. I've added the following:

- I've split all on_xxx logging functions into a class in the detail namespace and in a separate file, to make it clear that they're private.
- I've added an example on how to integrate spdlog.
- I've added constructors to connection taking a logger.
- I've decreased the default logging level to info, which I think is a reasonable default.
- I've made logging thread-safe.
- Added tests and docs.
- Deprecated the async_run overload taking a logger.

If you find this OK, I can merge it so you can build your read_fsm PR on top of this.

include/boost/redis/connection.hpp
```diff
@@ -755,25 +825,35 @@ class basic_connection {
       mpx_.cancel_on_conn_lost();
    }
 
-   template <class, class> friend struct detail::reader_op;
-   template <class, class> friend struct detail::writer_op;
-   template <class> friend struct detail::exec_op;
-   template <class, class> friend class detail::run_op;
+   // Used by both this class and connection
+   void set_run_logger(const logger& lgr, const config& cfg)
```
mzimbres: Shouldn't this be named `set_logger` since the logger can also be used in `async_exec`?

anarthal: What I meant with that name is "what async_run calls to set the logger". Note that constructors don't call this function. It is there because calling `basic_connection::async_run` from `connection::async_run`'s implementation caused warnings because of the deprecations. The assignment is special in the sense that it will discard the logging function and only use the level. I can probably find a better name for this.

PR #274, opened 2025-06-16
Adds a sans-io fsm to the read operation.

mzimbres:
@anarthal: Notes for the code review
   - Adds a fsm for read operation following your work on `exec_fsm`.
   - Log every event of the `reader_fsm` after it resumes, which improves log information content.
   - The logger interface has to be extended with types that are actually in the `detail` namespace, so I make fsm actions a public type, this is needed to interpret log messages. I would like to do this to the `exec_fsm` too once a log object becomes available for `async_exec`, so I did not change your code yet.
   - Cancellation hasn't been addresses yet for read operation. This will need more thought and a separate PR.

Example log for the subscriber example. Once all ops are refactored to use an fsm, the log will become very detailed and more helpful
```
(Boost.Redis) resolve results: 127.0.0.1:6379
(Boost.Redis) connected to 127.0.0.1:6379
(Boost.Redis) fsm action: (reader_action::type::setup_cancellation, 0, Success)
(Boost.Redis) fsm action: (reader_action::type::append_some, 0, Success)
(Boost.Redis) writer_op: 117 bytes written.
(Boost.Redis) fsm action: (reader_action::type::notify_push_receiver, 36, Success)
(Boost.Redis) fsm action: (reader_action::type::append_some, 0, Success)
(Boost.Redis) hello_op: Success
subscribe channel 1
^C
(Boost.Redis) ping_op (4): Operation canceled
(Boost.Redis) check_timeout_op (2): Operation canceled
(Boost.Redis) fsm action: (reader_action::type::cancel_run, 0, Success)
(Boost.Redis) fsm action: (reader_action::type::done, 0, Operation canceled)
(Boost.Redis) writer_op (3): connection is closed.
```


include/boost/redis/fsm_action.hpp
```diff
@@ -0,0 +1,60 @@
...
+      append_some,
+      /// Ask the reader operation to notify the push receiver that a push is available.
+      notify_push_receiver,
+      /// Cancel the connection's run operation
+      cancel_run,
+      /// Done with the reader operation
+      done,
+   };
+
+   reader_action(
```
anarthal: If you're going to only have one constructor, better make all members public, provide no constructors and make the type an aggregate. It'd make sense to make a constructor if you think of the action as a variant-like type - e.g. having a constructor taking only an `error_code` that sets the type to `done`. Up to you. If you choose that path, I'd make the constructor inline - I don't think the gains in compile time here compensate the runtime cost of separate compilation.

include/boost/redis/connection.hpp
```diff
@@ -206,85 +207,56 @@ struct writer_op {
...
-            res_ = conn_->mpx_.consume_next(ec);
+         logger_.on_fsm_resume(act);
 
-            if (ec) {
-               logger_.trace("reader_op (3)", ec);
-               conn_->cancel(operation::run);
-               self.complete(ec);
+         switch (act.type_) {
+            case reader_action::type::setup_cancellation:
+               self.reset_cancellation_state(asio::enable_total_cancellation());
```
anarthal: Maybe partial cancellation makes more sense here? In any case, we'll handle this later.

test/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ make_test(test_request)
 make_test(test_low_level_sync_sans_io)
 make_test(test_any_adapter)
 make_test(test_exec_fsm)
+make_test(test_reader_fsm)
```
anarthal, line 43: Since this test does not require a real server, please add it to test/Jamfile, too.

include/boost/redis/impl/exec_fsm.ipp
```diff
@@ -91,4 +87,22 @@ boost::redis::detail::exec_action boost::redis::detail::exec_fsm::resume(
    return exec_action{system::error_code()};
 }
 
-#endif
\ No newline at end of file
+#define BOOST_REDIS_EXEC_SWITCH_CASE(elem) \
+   case exec_action_type::elem: return "exec_action_type::" #elem
+
+auto to_string(exec_action_type t) noexcept -> char const*
```
anarthal: Do we need this in the public headers at this point, or is it only used in the tests? IIRC the exec FSM doesn't have access to the logger at this point.

test/test_reader_fsm.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+}  // namespace
+
+int main()
+{
+   test_push();
+   test_read_error();
+   test_parse_error();
+
```
anarthal, line 231: I'd add the following cases:

* An incomplete message is delivered (i.e. it's fragmented in two reads).
* The channel notification fails

include/boost/redis/impl/logger.ipp
```diff
@@ -89,19 +89,15 @@ void logger::on_write(system::error_code const& ec, std::string_view payload)
...
 
    write_prefix();
 
-   if (ec)
-      std::clog << "reader_op: " << ec.message();
-   else
-      std::clog << "reader_op: " << n << " bytes read.";
-
-   std::clog << std::endl;
+   std::clog << "fsm action: (" << to_string(action.type_) << ", " << action.push_size_ << ", "
```
anarthal: I've got the concern that we might be logging too much for the `level::info` level. It looks to me that most of this info really belongs to `level::debug`. However, fine-tuning this is probably better done once we merge the logger interface PR, because we can start moving logging to fsms (and hence testing it).

include/boost/redis/fsm_action.hpp
mzimbres: Ok. With your log PR I can also move this to the detail namespace too since reader_action won't then appear in the public API.

mzimbres:
I think I addressed all your points. Please have a look at it again. Are you about to get the log PR merged? I will wait and then rebase on top of your changes so that I can remove the usage of types in `detail` in the logger interface.

PR #275, opened 2025-06-22
Makes the reconnection loop retry on errors

### LINKED ISSUE #255: async_run exits after a resolve, connect or TLS handshake failure
After the first failure, `async_exec` returns, even if it's configured to retry indefinitely.

anarthal:
After an error is encountered during name resolution, connection establishment, SSL handshake, Redis hello, reading or writing, async_run will now keep retrying if the configuration requests it

close #255

### DIFF
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -292,29 +292,44 @@ public:
    : conn_{conn}
    { }
 
-   template <class Self>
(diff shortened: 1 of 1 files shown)


anarthal:
I've performed manual testing, both with and without TLS, verifying that retries are performed both on connection errors and on read errors (by killing the server in the middle of an operation).

PR #276, opened 2025-06-22
Migrates the documentation to Asciidoc/MrDocs

### LINKED ISSUE #247: The Doxygen reference seems to be empty
https://www.boost.org/doc/libs/develop/libs/redis/doc/html/classboost_1_1redis_1_1basic__connection.html

anarthal:
close #247


anarthal:
This is still a very early prototype with lots of things to solve.

mzimbres:
> What do you think? Do we move forward with this and accept the defects? Or do we set up a PyDocca/quickbook setup as I have in MySQL, which we know it works well?

I prefer to move directly to MrDocs and the reference looks great. Is it considered too early to use MrDocs or are they welcoming early adopters? I think we can go forward with this.

anarthal:
Preview: https://anarthal.github.io/boost-redis-mrdocs/redis/index.html

PR #277, opened 2025-06-25
Avoids suspending the reader fsm when a sync push send is possible

mzimbres:


anarthal:
I'm not convinced. What is the advantage of avoiding a fsm suspension? AFAIK suspending a FSM does not have any cost (it's not like suspending an Asio async operation, which has a very noticeable cost). On the other hand, creating and invoking a `std::function` is likely more costly than suspending a FSM.

mzimbres:
Regarding the importance I remember at least one use case from @ashtum where Boost.Redis was used to process thousands of pushes per second. This is just to say that performance is important.

mzimbres:
I let the idea ripe a bit more. For now I will move the `try_notify_push` callback to the multiplexer that is where I think it belongs to.

> Regarding code organization, performance aside, the approach closer to what we've been doing would be having a action::type::try_notify_push_receiver that triggers a channel::try_send and resumes the reader with whatever result the channel yielded.

If we take this direction I will also move notify_done out of the multiplexer into the reader_op so we have some consistency. I actually did this but then dropped the commit because it feels like it is the multiplexer role to unblock pending ops when they are ready to continue.

PR #278, opened 2025-06-27
Generic flat response implementation

D0zee:


D0zee:
What is left: 
- Implementation of `consume_one` function
- Tests / examples
- Documentation

include/boost/redis/response.hpp
```diff
@@ -34,6 +34,83 @@ using response = std::tuple<adapter::result<Ts>...>;
...
+   void push_back(const resp3::basic_node<String>& nd)
+   {
+      resp3::offset_node new_node;
+      new_node.data_type = nd.data_type;
+      new_node.aggregate_size = nd.aggregate_size;
+      new_node.depth = nd.depth;
+      new_node.offset = data_.size();
+      new_node.size = nd.value.size();
+
+      data_ += std::string{std::cbegin(nd.value), std::cend(nd.value)};
```
anarthal: This is making an avoidable copy - I'd suggest .append()

include/boost/redis/response.hpp
```diff
@@ -34,6 +34,83 @@ using response = std::tuple<adapter::result<Ts>...>;
...
+
+      view_.push_back(std::move(new_node));
+   }
+
+private:
+   resp3::node_view make_node_view(const resp3::offset_node& n)
+   {
+      return resp3::node_view{
```
anarthal: Named initializers are C++20

include/boost/redis/response.hpp
```diff
@@ -34,6 +34,83 @@ using response = std::tuple<adapter::result<Ts>...>;
...
 
+struct flat_response_impl {
+private:
+   class iterator {
+   public:
+      using value_type = resp3::node_view;
+      using difference_type = std::ptrdiff_t;
+      using pointer = void;
+      using reference = value_type;
+      using iterator_category = std::forward_iterator_tag;
```
anarthal: I think we can make this random access (although it will mean implementing more iterator methods)

include/boost/redis/response.hpp
```diff
@@ -34,6 +34,83 @@ using response = std::tuple<adapter::result<Ts>...>;
...
+      resp3::offset_node new_node;
+      new_node.data_type = nd.data_type;
+      new_node.aggregate_size = nd.aggregate_size;
+      new_node.depth = nd.depth;
+      new_node.offset = data_.size();
+      new_node.size = nd.value.size();
+
+      data_ += std::string{std::cbegin(nd.value), std::cend(nd.value)};
+
+      view_.push_back(std::move(new_node));
```
anarthal, line 93: Does this std::move add anything?

include/boost/redis/resp3/node.hpp
```diff
@@ -66,6 +66,14 @@ using node = basic_node<std::string>;
  */
 using node_view = basic_node<std::string_view>;
 
+/**
+ * TODO: documentation
+ */
+struct offset_node : node_view {
```
anarthal: Do we need to expose this as public? Looks like an implementation detail

include/boost/redis/response.hpp
```diff
@@ -34,6 +34,83 @@ using response = std::tuple<adapter::result<Ts>...>;
  */
 using generic_response = adapter::result<std::vector<resp3::node>>;
 
+struct flat_response_impl {
```
anarthal: I'd avoid the _impl suffix for public types - maybe flat_response_value?

anarthal (commented):
I've reviewed the flat_response_impl part. It looks what I'd expect. Marcelo is more suited to review the rest, since I don't really know how adapters work. Thanks for submitting the patch.

include/boost/redis/response.hpp
D0zee: I'm thinking about putting this structure into `boost::redis::response::detail`namespace as well

include/boost/redis/response.hpp
D0zee: Agreed!

include/boost/redis/response.hpp
D0zee: Frankly there is no difference between moving and copying in this case as the `offset_node` is trivially copyable. However I would prefer to move it anyway

include/boost/redis/response.hpp
D0zee: What I really think about is exposing method `data()` or something similar that returns `std::vector<node_view>` in `flat_response_value`. Thus if the user wants to iterate over node views he will call this method and use all benefits of std iterators

include/boost/redis/response.hpp
anarthal: I'd keep it public. `adapter::result` is just a `system::result` type, so the user needs to interact with `flat_response_impl` and know its API.

D0zee:
Hi @mzimbres, I'm going to look at suggested issues in details and implement them to enhance performance and make the structure more user-friendly. Thank you for your comments in the issue

include/boost/redis/response.hpp
```diff
@@ -34,6 +34,52 @@ using response = std::tuple<adapter::result<Ts>...>;
...
+   std::vector<resp3::offset_node>& view() { return view_; }
+
+   template <typename String>
+   void push_back(const resp3::basic_node<String>& nd)
+   {
+      resp3::offset_string offset_string;
+      offset_string.offset = data_.size();
+      offset_string.size = nd.value.size();
+
+      data_.append(nd.value.data());
```
mzimbres: `nd.value.data()` does not point to null terminated strings in general so you have to pass the size as well.

include/boost/redis/response.hpp
```diff
@@ -34,6 +34,52 @@ using response = std::tuple<adapter::result<Ts>...>;
...
+
+      offset_string.data = std::string_view{
+         data_.data() + offset_string.offset,
```
mzimbres: If the next call to `data_.append` from the previous line causes the string to reallocate, this pointer (`data_.data()`) will become invalid. To address this I have proposed in other comments to use the `notify_done` callback to traverse the flat response initializing the `string_view` of all offset strings. I have also considered delegating this step to the user, for example
```cpp
generic_flat_response resp;
co_await conn.async_exec(req, resp);
resp.set_views(); // Traverses all nodes setting the strings.
```
I am still not sure though as this is error prone.

D0zee: Oh, you are correct. Will fix it

D0zee: This comment was sorted out. Tested on `cpp20_subscriber`, it works as expected. If you have any other comments about `prepare_done` function or anything else, please let me know

example/cpp20_subscriber.cpp
mzimbres, line 1: I think we can just replace generic_response with generic_flat_response in all occurrences.

include/boost/redis/adapter/detail/adapters.hpp
```diff
@@ -138,6 +139,26 @@ void boost_redis_from_bulk(T& t, resp3::basic_node<String> const& node, system::
 
 //================================================
 
+using done_fn_type = std::function<void()>;
+
+template <typename T>
+auto prepare_done(T&) noexcept -> done_fn_type
```
mzimbres: I don't think there any need to type erase the return value at this point. We should delay it to the calling point where `prepare_done` is actually called.

D0zee: i was thinking about static function `prepare_done()` inside `any_adapter`that does nothing with a specialization for `generic_flat_response` that calls `set_view`. Would it work?

include/boost/redis/response.hpp
```diff
@@ -72,12 +124,22 @@ using generic_response = adapter::result<std::vector<resp3::node>>;
...
+void consume_one(generic_flat_response& r, system::error_code& ec);
+
 /**
  * @brief Throwing overload of `consume_one`.
  *
  * @param r The response to modify.
  */
-void consume_one(generic_response& r);
+template <typename Response>
+void consume_one(Response& r)
```
mzimbres: Can we avoid turning this function into a template? We can have two overloads, one for `generic_response` and one for `generic_flag_response` and move their implementations to the `.ipp` file. We want to avoid any increase in compilation times.

example/cpp20_subscriber.cpp
D0zee: It will be good test for API of new class because I already see some absent methods like `clear()`.

include/boost/redis/response.hpp
D0zee: The difference between two specializations is actually one string, that's why I made it template. However if I want to avoid any increase in compilation time, then you are right, I will fix it

include/boost/redis/adapter/any_adapter.hpp
```diff
@@ -34,10 +34,13 @@ namespace boost::redis {
...
 public:
-   using fn_type = std::function<void(std::size_t, resp3::node_view const&, system::error_code&)>;
+   using adapt_fn_type = std::function<
+      void(std::size_t, resp3::node_view const&, system::error_code&)>;
+   using done_fn_type = adapter::detail::done_fn_type;
 
    struct impl_t {
-      fn_type adapt_fn;
+      adapt_fn_type adapt_fn;
+      done_fn_type prepare_done_fn;
```
mzimbres: I would like to avoid the creation of two `std::function` for each request (i.e. `async_exec` calls) since each one represents a potential dynamic allocation. In this case however I think we can't do much better because inside `async_exec` we only have access to the type erased adapter i.e. `std::function` and therefore we can't call `adapter.set_done()` for example. I think the `prepare_done_fn` callback is not critical though since it is small and the dynamic allocation is likely to be optimized away with SOO (small object optimization). @anarthal What do you think?

D0zee:
```
sizeof(generic_flat_response*)=8
sizeof(generic_flat_response&)=64
sizeof([](){})=1
```

- If response type is `generic_flat_response` we will return lambda with captured reference (`[res]()mutable{}`). The size of this lambda is 64 bytes, `std::function` will definitely keep it on heap. I suggest to capture and pass the pointer`generic_flat_response*`. Its size is 8 bytes and in this case SOO will take place during creation of `std::function` in `any_adapter`. As a result we have no heap allocations in this case.

- If the response type is any other SOO will take a place by default due to the fact that size of empty lambda is 1 byte.

Source: https://stackoverflow.com/a/57049013

mzimbres: I have been thinking about this and there seems to be a simpler way of avoiding the extra `std::function` that I haven't considered earlier. First we define a special node value in `node.hpp`

```cpp
constexpr resp3::node_view done_node{resp3::type::invalid, -1, -1, {}};
```

then call the adapter with this node when the parser is done [here](https://github.com/boostorg/redis/blob/adf17f2b3b2eb119abe324f82dc93cea0a2d0065/include/boost/redis/resp3/parser.hpp#L97)

```cpp
template <class Adapter>
bool parse(resp3::parser& p, std::string_view const& msg, Adapter& adapter, system::error_code& ec)
{
   while (!p.done()) {
      ...
   }

   // ---> New adapter call here to inform parsing is finished.
   adapter(std::make_optional<resp3::parser::result>(done_node), system::error_code());

   return true;
}
```

Then call `set_views()` on the adapter when this node is detected [here](https://github.com/boostorg/redis/pull/278/files#diff-5d85143a4edd5e06ac41022496b7a05dd6c7f661b33eb380b23cf8f8dc92079bR204)

```cpp
template <>
class general_aggregate<result<flat_response_value>> {
private:
   result<flat_response_value>* result_;

public:
   template <class String>
   void operator()(resp3::basic_node<String> const& nd, system::error_code&)
   {
      // ---> Check whether done here.
      if (nd == done_node) {
         result_->value().set_views();
         return;
      }

      ...
   }
};
```

I totally missed this possibility earlier but it looks cleaner than adding a new `std::function`? Do you mind trying this?

D0zee: Yes, it looks much better and we don't have to care about heap allocations. I will try it!

D0zee: Am I correct that this case must be handled by others adapters as well?

mzimbres: Hi @D0zee, yes other adapters will have to handle this as well, for example
```cpp
if (nd == done_node)
    return;
```
I forgot to say this in my previous comment.

mzimbres: Hi again, there is still one problem to solve. The way I suggested above will cause the `set_view` to be called multiple times, once for each request in the pipeline. That is because the same flat response can be used to store the response of multiple requests. I am still not sure what is the simplest way to avoid that, it can be solved by adding state to `set_views()` so that it does not traverse what has already been traversed.

Or perhaps there is a way to wrap the adapters at other location such as [here](https://github.com/boostorg/redis/blob/develop/include/boost/redis/adapter/any_adapter.hpp#L44-L51)

```cpp
   template <class T>
   static auto create_impl(T& resp) -> impl_t
   {
      using namespace boost::redis::adapter;
      auto adapter = boost_redis_adapt(resp);
      std::size_t size = adapter.get_supported_response_size();
      return {std::move(adapter), size};
   }
```
or [here](https://github.com/boostorg/redis/blob/develop/include/boost/redis/impl/multiplexer.ipp#L22-L25)
```cpp
   adapter_ = [this, adapter](resp3::node_view const& nd, system::error_code& ec) {
      auto const i = req_->get_expected_responses() - remaining_responses_;
      adapter(i, nd, ec);
   };
```
so that only the last call to `adapter(done_node, ec)` triggers a `set_view()`. I think however this might be too complicated and perhaps it just simpler to let each call to set_view to traverse only what has not been set yet, as suggested above.

If I have time I will review all of this to see if there is any simplification possible.

mzimbres:
@anarthal There are two problems to solve

1. The `generic_flat_response` api
2. Where to construct the `string_views` from the offsets.

Number 1. is the easier part and I would like to have it equal to `generic_response`. The current PR is almost there but actually requires calling `resp.view()`. I am still not sure whether this is ok but it is something easy to change.

Number 2. is what is causing problems. I am still trying to understand what is cleaner from the design and performance point of view and in a way that doesn't break existing code. We have already two callbacks associated with `async_exec`, the adapter (and its wrapper) and the `set_done_callback`. The latter is the best place where we can call `flat_resp.set_view()` but type erasure is preventing that. Adding a new callback is a poor workaround which makes the code more messy. The third option is to let parser notify the adapter it is done.

I think the third option is the only clean and sound option because it makes sense to know when parsing is complete from the adapter. But once we have that the `set_done_callback` existence become perhaps unnecessary.

Given this complexity I think it would be simpler to split this PR in two. @D0zee works only on 1. and I will work on part 2. so he has a sane way of calling `set_views()` that does not mess with the code. After I am done he can rebase and it should work.

D0zee:
@mzimbres Let me please know when your part is ready. I think I will finish this PR tomorrow and on weekends

mzimbres:
@D0zee I will. I am currently finishing [this](https://github.com/boostorg/redis/pull/283) PR and after it is merged I will investigate how to solve this one.

anarthal:
>We have already two callbacks associated with async_exec, the adapter (and its wrapper) and the set_done_callback. 

Note that the done callback currently belongs to the I/O world (it interacts with a channel) rather than the parser world. IMO long-term the callback should be replaced by a reader action (as I think was your intention according to [this comment](https://github.com/boostorg/redis/blob/adf17f2b3b2eb119abe324f82dc93cea0a2d0065/include/boost/redis/impl/reader_fsm.ipp#L59-L64)).

Semantically, having adapters support a "done" callback looks sound to me. But as you said, there is the type erasing issue.

If you want, I can try to write a type-erased adapter type that encapsulates both functions, something like:

```cpp
class any_adapter_impl {
    // stores the underlying adapter, akin to what std::function does for functions
public:
    void on_node(std::size_t, resp3::node_view const&, system::error_code&);
    void on_finished();
};
```

Then you can use this type in `parser`.

D0zee:
Hi @mzimbres, I've resolved the comments above and added api methods required for tests/examples. However I realized that examples don't work because we have to do the same trick as with `set_done_callback` in `multiplexer` [here](https://github.com/boostorg/redis/blob/adf17f2b3b2eb119abe324f82dc93cea0a2d0065/include/boost/redis/detail/multiplexer.hpp#L174). 

With adapter extension mentioned below we will be able to get rid of `prepare_done` callback in `any_adapter` and trick I mentioned below will be not needed. Looking forward to the implementation, it would unblock me. Thank you in advance!

mzimbres:
@D0zee Thanks. We are working on the prerequisites for this PR and will notify you when it is finished so we can proceed with this one.

include/boost/redis/adapter/detail/adapters.hpp
```diff
@@ -174,6 +174,41 @@ class general_aggregate {
...
+public:
+   explicit general_aggregate(result<flat_response_value>* c = nullptr)
+   : result_(c)
+   { }
+
+   void on_init() { }
+   void on_done()
+   {
+      if (result_->has_value()) {
+         result_->value().set_view();
```
mzimbres, line 191: If a request contains multiple commands, `on_done()` and by consequence `set_view()` will be called multiple times, although it should be called only once on the last `on_done()` call. I have to review the code again to understand whether we know upfront how many times `on_done()` will be called. Traversing the vector of nodes every time is prohibitively expensive.

This might get more complex when reading server pushes with async_receive since we don't really know how many we are expecting.

D0zee: Actually I have not thought about it, I will consider this case as well

mzimbres:
Hi again, regarding the problem of calling `set_view()` only at the last `on_done()` parse event: This is probably simple to solve if if we pass [the number of expected responses](https://github.com/boostorg/redis/blob/da2f0101d01cf45b848a11a8936acf1d47cc5b4e/include/boost/redis/request.hpp#L87) to the adapter. However, when dealing with server pushes we don't know how many will be received upfront so the library would have to call `set_view()` right before resuming `async_receive()`, which would be probably [here](https://github.com/boostorg/redis/blob/da2f0101d01cf45b848a11a8936acf1d47cc5b4e/include/boost/redis/connection.hpp#L294), but we can't since we only have a type erased push-receiver adapter. 

I think I also oversaw this problem in the first stages of this PR that used the set_done_callback, that only works for request/response but not for server pushes.

In summary, it might be better to delegate the `set_view()` call to the user instead of finding a way to make the connection or adapter call it on behalf thereof. If we can agree on that I would like to revert the changes to the examples that replace all `generic_response` occurrences with `generic_flat_response`. We should use only one example for now and comment on the need to call `set_view()`.

Thoughts?

anarthal:
Pushing the responsibility to the user seems hostile. The library seems the right place to do it.

If the `on_finished` adapter callback does not suit this use case, what's its use then? Shouldn't it be either removed or replaced by something that does suit (e.g. `on_all_finished`)?

On the other hand, why was the idea of creating the `string_view` objects on demand (i.e. a custom range) ruled out? Has the "less efficient" call been backed by any actual data?

criatura2:
After some days thinking I believe there is a good solution that just requires making `add_node` and `set_view` smarter.

  1. Add a new field `std::size_t pos_ = 0` to `flat_response_value`.

  2. Let `set_view` use `pos_` to set the views incrementally when it gets called by `on_done`

```cpp
void set_view()
 {
    for (; pos_ < view_.size(); ++pos_) {
       auto& v = view_.at(pos_).value;
       v.data = std::string_view{data_.data() + v.offset, v.size};
    }
 }
```

  3. Move the current `add_node` implementation to `add_node_impl` and let `add_node` detect any memory allocations

```cpp
void add_node(resp3::basic_node<String> const& nd)
{
  auto capacity_before = data_.capacity();
  add_node_impl(nd);
  auto capacity_after = data_.capacity();

  if (capacity_after > capacity_before)
    pos_ = 0;
}
```

That means, every time `on_done` calls `set_view` it will only traverse nodes that haven't been set yet. When a memory reallocation is detected `pos_` is set to `0` so that all nodes get reset on the next `on_done` call. Reallocation however will either not occur if the user reserves memory upfront or only a couple of times at program startup since responses are reused, meaning that asymptotically it will not allocate any memory.

@D0zee Thanks for keeping up so far, this PR has been very helpful in detecting open points in the original design. If you don't have the time to implement the suggestion above I can take over it, I think however this is not much work. Thanks.

criatura2:
> @criatura2 Thank you for your proposal, but I think the similar problem was discussed at some moment of time in this PR and we denied that. The intention is to call set_view() only once to be efficient.

What is not ok is to traverse the whole vector for each `on_done` call, which is a O(n^2) solution. It is ok however to traverse only the newly added items, which is an amortized O(n) solution. And as said previously, since re-allocations are very rare or non-existent, the solution would be close to optimal.

criatura2:
> I suspect your solution to be O(n * log(n)) where n is the number of response strings in generic_flat_response, because of resetting pos_ when reallocation happens.

If the response is reused such as [here](https://github.com/boostorg/redis/blob/2133ed747b246c5c9278c1c88a334d10ad4a0036/example/cpp20_subscriber.cpp#L57) only one read will be `O(n * log(n))` after that it gets `O(n)`. Also, something like

```cpp
generic_flat_response resp;
resp.value().reserve(some_number);
```
Will bring reallocations to negligible levels.

D0zee:
@anarthal yes, exactly

mzimbres:
> I want to understand why this can't be done after parsing is finished - is it because of pushes?

Yes. The default size of the channel used to deliver pushes is 256, that means `on_done()` can be called that many times and you never know which one is the last call so that `set_view()` can be called. We might want to call `set_view` right after `try_send` and just before `async_send` [here](https://github.com/boostorg/redis/blob/0cf2441ed2b81506f4b6f626d01ea43785ba31a9/include/boost/redis/connection.hpp#L295) but that does not work either because that would be a race condition i.e. the receiver side might suspend allowing a further push to be processed after `set_view` has been called.

D0zee:
Hi @mzimbres, didn't know that account is yours! I need to think about it one more time. Sorry I was busy this week as well. Thank you for bearing with me!

D0zee:
> If you don't have the time to implement the suggestion above I can take over it, I think however this is not much work. Thanks.


Hi @mzimbres, can you please take over that change? I understand that I've been extremely busy recently and it stops this PR to be merged.

mzimbres:
> Hi @mzimbres, can you please take over that change? I understand that I've been extremely busy recently and it stops this PR to be merged.

Hi @D0zee, no problem I will pick your commits to properly attribute your the work you did. I will also pull you in the PR when it is ready. Thanks.

D0zee:
Hi @mzimbres, nvladimirov.work@gmail.com. Thank you!

PR #283, opened 2025-07-10
Removes async_append_some

mzimbres:
@anarthal Before I start investigating how to proceed with @D0zee PR I would like to merge this one which simplifies how the read buffer is used.

### DIFF
--- a/include/boost/redis/config.hpp
+++ b/include/boost/redis/config.hpp
@@ -88,7 +88,7 @@ struct config {
     */
    std::chrono::steady_clock::duration reconnect_wait_interval = std::chrono::seconds{1};
 
-   /** @brief Maximum size of the read-buffer in bytes.
+   /** @brief Maximum size of the socket read-buffer in bytes.
     *  
     *  Sets a limit on how much data is allowed to be read into the
     *  read buffer. It can be used to prevent DDOS.
@@ -98,8 +98,9 @@ struct config {
    /** @brief read_buffer_append_size
     *
     * The size by which the read buffer grows when more space is
-    * needed. There is no need to set this too high because memory is
-    * reused and the growth will tend to zero.
+    * needed. This can help avoiding some memory allocations. Once the
+    * maximum size is reached no more memory allocations are made
+    * since the buffer is reused.
     */
    std::size_t read_buffer_append_size = 4096;
 };
--- a/include/boost/redis/config.hpp
+++ b/include/boost/redis/config.hpp
@@ -88,12 +88,20 @@ struct config {
     */
    std::chrono::steady_clock::duration reconnect_wait_interval = std::chrono::seconds{1};
 
-   /** @brief Maximum size of a socket read, in bytes.
+   /** @brief Maximum size of the read-buffer in bytes.
     *  
     *  Sets a limit on how much data is allowed to be read into the
     *  read buffer. It can be used to prevent DDOS.
     */
    std::size_t max_read_size = (std::numeric_limits<std::size_t>::max)();
+
+   /** @brief read_buffer_append_size
+    *
+    * The size by which the read buffer grows when more space is
+    * needed. There is no need to set this too high because memory is
+    * reused and the growth will tend to zero.
+    */
+   std::size_t read_buffer_append_size = 4096;
 };
 
 }  // namespace boost::redis
--- a/include/boost/redis/detail/multiplexer.hpp
+++ b/include/boost/redis/detail/multiplexer.hpp
@@ -8,13 +8,13 @@
 #define BOOST_REDIS_MULTIPLEXER_HPP
 
 #include <boost/redis/adapter/adapt.hpp>
-#include <boost/redis/adapter/any_adapter.hpp>
-#include <boost/redis/config.hpp>
-#include <boost/redis/operation.hpp>
+#include <boost/redis/detail/read_buffer.hpp>
+#include <boost/redis/resp3/node.hpp>
+#include <boost/redis/resp3/parser.hpp>
 #include <boost/redis/resp3/type.hpp>
 #include <boost/redis/usage.hpp>
 
-#include <boost/asio/experimental/channel.hpp>
+#include <boost/system.hpp>
 
 #include <algorithm>
 #include <deque>
@@ -32,7 +32,8 @@ namespace detail {
 
 using tribool = std::optional<bool>;
 
-struct multiplexer {
+class multiplexer {
+public:
    using adapter_type = std::function<void(resp3::node_view const&, system::error_code&)>;
    using pipeline_adapter_type = std::function<
       void(std::size_t, resp3::node_view const&, system::error_code&)>;
@@ -127,7 +128,8 @@ struct multiplexer {
    // If the tribool contains no value more data is needed, otherwise
    // if the value is true the message consumed is a push.
    [[nodiscard]]
-   auto consume_next(system::error_code& ec) -> std::pair<tribool, std::size_t>;
+   auto consume_next(std::string_view data, system::error_code& ec)
+      -> std::pair<tribool, std::size_t>;
 
    auto add(std::shared_ptr<elem> const& ptr) -> void;
    auto reset() -> void;
@@ -156,18 +158,6 @@ struct multiplexer {
       return std::string_view{write_buffer_};
    }
 
-   [[nodiscard]]
-   auto get_read_buffer() noexcept -> std::string&
-   {
-      return read_buffer_;
-   }
-
-   [[nodiscard]]
-   auto get_read_buffer() const noexcept -> std::string const&
-   {
-      return read_buffer_;
-   }
-
    // TODO: Change signature to receive an adapter instead of a
    // response.
    template <class Response>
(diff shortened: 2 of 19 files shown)


include/boost/redis/detail/read_buffer.hpp
```diff
@@ -0,0 +1,46 @@
...
+public:
+   void prepare_append(
+      std::size_t append_size,
+      std::size_t max_buffer_size,
+      system::error_code& ec);
+
+   void commit_append(std::size_t read_size);
+
+   [[nodiscard]]
+   auto get_append_buffer() noexcept -> std::pair<char*, std::size_t>;
```
anarthal: Consider replacing by `boost::span<char>`

include/boost/redis/config.hpp
```diff
@@ -88,12 +88,20 @@ struct config {
...
    std::chrono::steady_clock::duration reconnect_wait_interval = std::chrono::seconds{1};
 
-   /** @brief Maximum size of a socket read, in bytes.
+   /** @brief Maximum size of the read-buffer in bytes.
     *  
     *  Sets a limit on how much data is allowed to be read into the
     *  read buffer. It can be used to prevent DDOS.
     */
-   std::size_t max_read_size = (std::numeric_limits<std::size_t>::max)();
+   std::size_t max_read_buffer_size = (std::numeric_limits<std::size_t>::max)();
```
anarthal: This is a braking change, is there a possibility to maintain the old name?

include/boost/redis/detail/read_buffer.hpp
```diff
@@ -0,0 +1,46 @@
...
+#define BOOST_REDIS_READ_BUFFER_HPP
+
+#include <cstddef>
+#include <string>
+#include <string_view>
+#include <utility>
+
+namespace boost::redis::detail {
+
+class read_buffer {
```
anarthal, line 20: I suggest adding a test_read_buffer unit test that covers these functions, even if they're small

include/boost/redis/detail/read_buffer.hpp
```diff
@@ -0,0 +1,46 @@
...
+public:
+   void prepare_append(
```
anarthal: Suggest returning the error code with [[nodiscard]]

include/boost/redis/detail/reader_fsm.hpp
```diff
@@ -41,8 +47,20 @@ class reader_fsm {
...
+   {
+      return read_buffer_.get_append_buffer();
+   }
+
 private:
    int resume_point_{0};
+   read_buffer read_buffer_;
+   config cfg_;
    action action_after_resume_;
    action::type next_read_type_ = action::type::append_some;
```
anarthal, line 49: Why do we need a needs_more action and a next_read_type_? It looks like it has the same semantics as append_some.

append_some should likely be renamed to read_some

include/boost/redis/detail/reader_fsm.hpp
```diff
@@ -41,8 +47,20 @@ class reader_fsm {
...
+
+   [[nodiscard]]
+   auto get_append_buffer() noexcept
+   {
+      return read_buffer_.get_append_buffer();
+   }
+
 private:
    int resume_point_{0};
+   read_buffer read_buffer_;
```
anarthal: I'd make read_buffer_ a property of the connection, and pass here a pointer. This way, FSMs are kept non-owning. This way, you don't need reset(), and can just re-create the FSM every time, making it impossible to create bugs by forgetting members in reset().

include/boost/redis/detail/read_buffer.hpp
```diff
@@ -0,0 +1,46 @@
...
+   void commit_append(std::size_t read_size);
+
+   [[nodiscard]]
+   auto get_append_buffer() noexcept -> std::pair<char*, std::size_t>;
+
+   [[nodiscard]]
+   auto get_committed_buffer() const noexcept -> std::string_view;
+
+   [[nodiscard]]
+   auto get_committed_size() const noexcept -> std::size_t;
```
anarthal, line 42: Isn't this just `get_committed_buffer().size()`? Do we need it?

include/boost/redis/detail/reader_fsm.hpp
```diff
@@ -41,8 +47,20 @@ class reader_fsm {
...
+   config cfg_;
    action action_after_resume_;
    action::type next_read_type_ = action::type::append_some;
    multiplexer* mpx_ = nullptr;
```
anarthal, line 50: As written, this yields to a dangling pointer after a connection is moved. I suggest wrapping every member in `basic_connection` into a state object, and placing this into the heap via a `unique_ptr` so they're guaranteed to have stable addresses.

include/boost/redis/detail/multiplexer.hpp
```diff
@@ -8,13 +8,13 @@
...
-#include <boost/redis/config.hpp>
-#include <boost/redis/operation.hpp>
+#include <boost/redis/detail/read_buffer.hpp>
+#include <boost/redis/resp3/node.hpp>
+#include <boost/redis/resp3/parser.hpp>
 #include <boost/redis/resp3/type.hpp>
 #include <boost/redis/usage.hpp>
 
-#include <boost/asio/experimental/channel.hpp>
+#include <boost/system.hpp>
```
anarthal: Please avoid whole-lib includes

include/boost/redis/impl/multiplexer.ipp
```diff
@@ -93,18 +93,16 @@ std::pair<tribool, std::size_t> multiplexer::consume_next(system::error_code& ec
    //    2. On a new message, in which case we have to determine
    //       whether the next messag is a push or a response.
    //
+
+   BOOST_ASSERT(!data.empty());
    if (!on_push_)  // Prepare for new message.
-      on_push_ = is_next_push();
+      on_push_ = is_next_push(data);
```
anarthal, line 99: This is likely not something to fix in this PR, but isn't the multiplexer violating point 1. in the comment above? While we're parsing something that's not a push and we run out of data, on_push_ will be false and we will look at the message's first byte even if we shouldn't.

include/boost/redis/impl/multiplexer.ipp
```diff
@@ -222,35 +237,29 @@ auto multiplexer::cancel_on_conn_lost() -> std::size_t
...
 }
 
-std::size_t multiplexer::on_finish_parsing(bool is_push)
+void multiplexer::commit_usage(bool is_push, std::size_t size)
 {
    if (is_push) {
       usage_.pushes_received += 1;
-      usage_.push_bytes_received += parser_.get_consumed();
+      usage_.push_bytes_received += size;
+      on_push_ = false;
```
anarthal, line 245: commit_usage setting on_push looks semantically surprising

include/boost/redis/detail/read_buffer.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+   void clear();
+
+   void consume(std::size_t size);
+
+private:
+   std::string buffer_;
```
anarthal: Might be better represented as `std::vector<char>`, instead. SBO here hurts more than it helps, and the contents are semantically closer to an array of bytes than a string (i.e. traits here is irrelevant).

test/common.hpp
```diff
@@ -34,3 +35,5 @@ void run(
    boost::redis::config cfg = make_test_config(),
    boost::system::error_code ec = boost::asio::error::operation_aborted,
    boost::redis::operation op = boost::redis::operation::receive);
+
+void append_read_data(boost::redis::detail::reader_fsm& fsm, std::string const& data);
```
anarthal: string_view?

test/common.cpp
```diff
@@ -69,3 +68,9 @@ void run_coroutine_test(net::awaitable<void> op, std::chrono::steady_clock::dura
       throw std::runtime_error("Coroutine test did not finish");
 }
 #endif  // BOOST_ASIO_HAS_CO_AWAIT
+
+void append_read_data(boost::redis::detail::reader_fsm& fsm, std::string const& data)
+{
+   auto const buffer = fsm.get_append_buffer();
```
anarthal: maybe BOOST_ASSERT that we don't exceed the buffer size?

include/boost/redis/detail/reader_fsm.hpp
mzimbres: needs_more is useful in some situations
1. Testing: I know that the fsm is being suspended exactly [here](https://github.com/boostorg/redis/blob/adf17f2b3b2eb119abe324f82dc93cea0a2d0065/include/boost/redis/impl/reader_fsm.ipp#L47).
2. Logging: Most responses will be contained entirely in the buffer when parsing starts. I would however like to see in the log when parsing has to be suspended because the current message is incomplete and more is needed.

> append_some should likely be renamed to read_some

That is reasonable, I can change it.

include/boost/redis/detail/read_buffer.hpp
mzimbres: Yeah, but it avoids creating a `string_view` just to read its size, which is a readly available member of the `read_buffer`.

include/boost/redis/impl/multiplexer.ipp
mzimbres: Yeah, it is not the correct place but it works well because `commit_usage` is called only once when finished with an individual response. This has to be moved elsewhere.

include/boost/redis/detail/read_buffer.hpp
mzimbres: This is not public, so I avoided pulling in another boost dependency.

include/boost/redis/detail/reader_fsm.hpp
mzimbres: Yeah, also considered this but did not go so far because I had change so much code already.

include/boost/redis/impl/multiplexer.ipp
mzimbres: I have to review this again and test it. I remember it being trick see comments in `is_next_push`.

mzimbres:
@anarthal I think I addressed all your comments. Your other PR will fix the remaining problems.

anarthal:
Looks good to me. I'll merge mine after you merge this.

PR #285, opened 2025-07-14
Makes all objects in connection have a stable address

anarthal:
Moves read_buffer memory reservation to the connection's constructor
Makes read_buffer memory reservation size be a power of 2


mzimbres:
Hi Ruben, I am going to address those comments in the way you suggested. Just don't have much time now. I agree what you propose here is good (and common) practice and does not have any bad side effects. I support further development here, thanks.

anarthal:
Great. I will do some further fixes here and post it for review again.

I'm going to be out of office during the weekend too, so don't worry.

anarthal:
This is now ready @mzimbres. I've made all composed operations use `connection_impl`, and moved some of the functionality required by them there. I've also moved the reserve() call to the connection's constructor.

include/boost/redis/connection.hpp
```diff
@@ -382,16 +503,8 @@ class basic_connection {
...
-   , writer_timer_{ex}
-   , reconnect_timer_{ex}
-   , receive_channel_{ex, 256}
-   , health_checker_{ex}
-   , logger_{std::move(lgr)}
-   {
-      set_receive_response(ignore);
-      writer_timer_.expires_at((std::chrono::steady_clock::time_point::max)());
-   }
+   : impl_(new detail::connection_impl<Executor>{std::move(ex), std::move(ctx), std::move(lgr)})
```
mzimbres: `make_unique`?

anarthal: What would be the benefit? When using a unique_ptr as a variable, it saves you from spelling the type twice:

```cpp
std::unique_ptr<int> myint ( new int{42} );
auto myint = std::make_unique<int>(42);
```

But here it seems to add nothing:

```
impl_(new detail::connection_impl<Executor>{std::move(ex), std::move(ctx), std::move(lgr)})
impl_(std::make_unique<detail::connection_impl<Executor>>{std::move(ex), std::move(ctx), std::move(lgr)})
```

Note that `make_unique` doesn't improve efficiency as `make_shared` does.

mzimbres: Correct, in this case I would only use it to avoid direct usages of the `new` operator.

include/boost/redis/connection.hpp
```diff
@@ -58,54 +58,157 @@ namespace boost::redis {
...
+   config cfg_;
+   multiplexer mpx_;
+   connection_logger logger_;
+   read_buffer read_buffer_;
+
+   using executor_type = Executor;
+
+   executor_type get_executor() noexcept { return writer_timer_.get_executor(); }
+
+   struct exec_op {
```
mzimbres, line 95: Why is `exec_op` embedded in the `connection_impl` while other op implementations are outside? It uses
`connection_impl* obj_ ;` instead of `connection_impl<Executor>* obj_;`

anarthal: Because it's used in connection_impl::async_exec, which is in turn needed to do the HELLO and the pings. The other ops are not.

connection_impl without template arguments in the context of the class is a shorthand for connection_impl<Executor>(see https://stackoverflow.com/questions/12810630/how-do-i-refer-to-a-class-template-as-a-template-from-within-its-own-class-def)

anarthal: My long-term plan is to place all ops within connection_impl once they're small enough and using FSMs, but I didn't want to change too much.

mzimbres (approved):

PR #286, opened 2025-07-25
Add parse event init, node and done.

mzimbres:
This PR implements parse events and is a prerequisite to the `generic_flat_response` from @D0zee.

### DIFF
--- a/include/boost/redis/adapter/any_adapter.hpp
+++ b/include/boost/redis/adapter/any_adapter.hpp
@@ -19,18 +19,6 @@
 
 namespace boost::redis {
 
-/** @brief Parse events that an adapter must support.
- */
-enum class parse_event
-{
-   /// Called before the parser starts processing data
-   init,
-   /// Called for each and every node of RESP3 data
-   node,
-   /// Called when done processing a complete RESP3 message
-   done
-};
-
 /** @brief A type-erased reference to a response.
  *
  *  A type-erased response adapter. It can be executed using @ref connection::async_exec.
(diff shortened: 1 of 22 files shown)


include/boost/redis/adapter/any_adapter.hpp
```diff
@@ -29,46 +41,63 @@ namespace boost::redis {
  *
  *  @code
  *      co_await conn.async_exec(req, resp);
- *      co_await conn.async_exec(req, any_response(resp));
+ *      co_await conn.async_exec(req, any_adapter(...));
  *  @endcode
  */
-class any_adapter {
+using any_adapter = std::function<void(parse_event, resp3::node_view const&, system::error_code&)>;
```
anarthal: I don't think this is the right interface for `any_adapter`, because

* There's now a discrepancy between what adapters support (3 functions with different parameters) and what `any_adapter` is (a single function with an enum)
* Users can't create `any_adapter` objects, since the only practical way is using `make_any_adapter`, which is in detail.

Instead of this, I'd make the `std::function` and the enum an implementation detail, like this:

```cpp
class any_adapter {
   enum class parse_event
   {
      /// Called before the parser starts processing data
      init,
      /// Called for each and every node of RESP3 data
      node,
      /// Called when done processing a complete RESP3 message
      done
   };

   using fn_type = std::function<void(parse_event, resp3::node_view const*, system::error_code*)>;
   fn_type impl_;

   template <class Response>
   static fn_type create_impl(Response& r)
   {
      using namespace boost::redis::adapter;

      return [adapter = boost_redis_adapt(
                 r)](parse_event ev, resp3::node_view const* nd, system::error_code* ec) mutable {
         switch (ev) {
            case parse_event::init: adapter.on_init(); break;
            case parse_event::node: adapter.on_node(*nd, *ec); break;
            case parse_event::done: adapter.on_done(); break;
         }
      };
   }

public:
   template <class T, class = std::enable_if_t<!std::is_same_v<T, any_adapter>>>
   explicit any_adapter(T& resp)
   : impl_(create_impl(resp))
   { }

   void on_init() { impl_(parse_event::init, nullptr, nullptr); };

   void on_done() { impl_(parse_event::done, nullptr, nullptr); };

   void on_node(resp3::node_view const& nd, system::error_code& ec)
   {
      impl_(parse_event::node, &nd, &ec);
   };
};
```

include/boost/redis/adapter/any_adapter.hpp
```diff
@@ -29,46 +41,63 @@ namespace boost::redis {
...
+   void on_node(resp3::node_view const& nd, system::error_code& ec)
+   {
+      adapter_(parse_event::node, nd, ec);
+   };
+
+   auto get_remaining_responses() const -> std::size_t { return expected_responses_; }
+
+private:
+   any_adapter adapter_;
+   std::size_t expected_responses_ = 0;
```
anarthal: I don't think this belongs here. You already maintain this information in `multiplexer::elem::remaining_responses_`. You can use this variable and go.

include/boost/redis/adapter/detail/response_traits.hpp
```diff
@@ -21,26 +22,6 @@
...
-      switch (nd.data_type) {
-         case resp3::type::simple_error: ec = redis::error::resp3_simple_error; break;
-         case resp3::type::blob_error:   ec = redis::error::resp3_blob_error; break;
-         case resp3::type::null:         ec = redis::error::resp3_null; break;
-         default:                        ;
-      }
-   }
-
-   [[nodiscard]]
-   auto get_supported_response_size() const noexcept
```
anarthal, line 38: It's a shame that we're getting rid of this because I think it helped debugging. Still, I think the implementation wasn't very complete, so I'm good with removing it for now. I'd open an issue to implement this better. Something such that if I type:


```cpp
request req;
req.push("ping", "abc");
req.push("ping", "def");

response<std::string> res;

co_await conn.async_exec(req, res);
```

Gets a proper diagnostic (I think right now it'll get a `std::out_of_range` from `vector::at`).

test/test_any_adapter.cpp
```diff
@@ -24,23 +25,23 @@ BOOST_AUTO_TEST_CASE(any_adapter_response_types)
    response<int, std::string> r2;
    generic_response r3;
 
-   BOOST_CHECK_NO_THROW(any_adapter{r1});
-   BOOST_CHECK_NO_THROW(any_adapter{r2});
-   BOOST_CHECK_NO_THROW(any_adapter{r3});
-   BOOST_CHECK_NO_THROW(any_adapter{ignore});
+   BOOST_CHECK_NO_THROW(make_any_adapter(r1));
```
anarthal: I think that these should work as they are (without the make_).

include/boost/redis/adapter/detail/response_traits.hpp
mzimbres: Exactly, the library should issue good diagnostics about such errors, for example

> Response object cannot accommodate the responses to all commands in the request.

Allowing `at()` to throw is not ok IMO.

mzimbres: Offloaded in https://github.com/boostorg/redis/issues/288

mzimbres:
@anarthal I think all concerns were addressed, please have a look at it again.

include/boost/redis/adapter/any_adapter.hpp
```diff
@@ -29,44 +29,69 @@ namespace boost::redis {
...
- *      co_await conn.async_exec(req, any_response(resp));
+ *      co_await conn.async_exec(req, any_adapter(...));
  *  @endcode
  */
 class any_adapter {
 public:
-   using fn_type = std::function<void(std::size_t, resp3::node_view const&, system::error_code&)>;
+   /** @brief Parse events that an adapter must support.
+    */
+   enum class parse_event
```
anarthal, line 39: What's the point of making `parse_event`, `impl_type` and the constructor from `impl_type` public and the constructor from the adapter private? `parse_event` and `impl_type` look like private things, while as a user, I want to be able to construct `any_adapter` objects from typed adapters.

include/boost/redis/adapter/any_adapter.hpp
```diff
@@ -29,44 +29,69 @@ namespace boost::redis {
...
+   /// Contructs from a type erased adaper
+   any_adapter(impl_type fn = [](parse_event, resp3::node_view const&, system::error_code&) { })
+   : impl_{std::move(fn)}
+   { }
 
-   template <class T>
-   static auto create_impl(T& resp) -> impl_t
+   // No need to document this as it is only used internally.
+   template <class Adapter>
+   any_adapter(Adapter adapter)
```
anarthal: Making this an `Adapter` (vs a `Response&`) is a breaking API change. To be honest, it makes sense for `any_adapter` to be constructible from an `Adapter` rather than a `Response&`, though. It is also true that all uses I've found end up calling `boost_redis_adapt`, so this feels like making syntax a little bit more convoluted. Is there any chance to make `any_adapter(response)` keep working?

anarthal:
@mzimbres see my comments. I still have concerns about `any_adapter`'s current interface.

include/boost/redis/adapter/any_adapter.hpp
mzimbres: After the Boost review customization for your own response was to based on providing `boost_redis_adapt()` that would be found via ADL. No `async_exec` overload that took adapters were expose.

Now that adapters are exposed in the async_exec overload it does not make sense to provide `boost_redis_adapt` for a custom response since the user can just construct an `any_adapter` himself by passing an `impl_type`. To enable that we have to make `impl_type` and `parse_event` public types.

anarthal: Why can't the user just write an adapter like this?

```cpp
class my_adapter {
  // impl
public:
    void on_init() { /* */ }
    void on_node(resp3::node_view const&, system::error_code&) { /* */ }
    void on_finish {}
};

// Now use it
conn.async_exec(req, any_adapter(my_adapter(/* whatever */));
```

mzimbres: This is supported in this PR: https://github.com/boostorg/redis/blob/2decc7b5120dd6f2f06a0b2ca93c016810e3aed7/include/boost/redis/adapter/any_adapter.hpp#L58-L71

include/boost/redis/adapter/any_adapter.hpp
mzimbres: I thought this was a small enough breaking change that wouldn't hurt anyone.

include/boost/redis/adapter/any_adapter.hpp
anarthal: It says "it is only used internally, so not documented"

include/boost/redis/adapter/any_adapter.hpp
anarthal: It will break at least the user who originally requested `any_adapter`

anarthal (approved):
This looks much better. If I have the time, I'd like to add a constructor to any_adapter from an `Adapter&&`, but that can be added later. Feel free to merge the PR.

PR #289, opened 2025-07-27
Fixes an exception when parsing intercalated errors and success msgs into generic_response

### LINKED ISSUE #287: generic_response's adapter throws an exception when receiving an error in an intermediate command
The following throws when it shouldn't:

```cpp
   config cfg;

   request req;
   req.push("PING", "hello");
   req.push("set", "mykey");
   req.push("get", "mykey");

   boost::redis::generic_response resp;

   asio::io_context ioc;
   connection conn{ioc};

   conn.async_run(cfg, asio::detached);

   conn.async_exec(req, resp, [&](auto, auto) {
      std::cout << "on async_exec" << std::endl;
      conn.cancel();
   });

   ioc.run();
```

The server will respond with, at least, with nodes like the following: [success, error, success].

This causes the adapter to behave incorrectly in [these lines](https://github.com/boostorg/redis/blob/20ab2c7e70ad9c9bd696acc69644a85c645c77bc/include/boost/redis/adapter/detail/adapters.hpp#L142-L171). The error node will emplace an error response in the result, and the next success node will invoke `result::value()`, throwing an exception.

I don't think this is super urgent but should be kept track of. If I have the time to refactor adapters at some point, I'll fix it.

anarthal:
close #287

### DIFF
--- a/test/test_conn_exec_error.cpp
+++ b/test/test_conn_exec_error.cpp
@@ -292,4 +292,38 @@ BOOST_AUTO_TEST_CASE(subscriber_wrong_syntax)
    BOOST_TEST(c3_called);
 }
 
+BOOST_AUTO_TEST_CASE(issue_287_generic_response_error_then_success)
+{
+   // Setup
+   auto cfg = make_test_config();
+   request req;
+   req.push("PING", "hello");
+   req.push("set", "mykey");  // This command has a missing argument and will cause an error
+   req.push("get", "mykey");  // This one is okay
+   generic_response resp;
+
+   // I/O objects
+   net::io_context ioc;
+   connection conn{ioc};
+   bool run_finished = false, exec_finished = false;
+
+   conn.async_run(cfg, [&](error_code ec) {
+      BOOST_TEST(ec == net::error::operation_aborted);
+      run_finished = true;
+   });
+
+   conn.async_exec(req, resp, [&](error_code ec, std::size_t) {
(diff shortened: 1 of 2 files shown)


test/test_conn_exec_error.cpp
```diff
@@ -292,4 +292,38 @@ BOOST_AUTO_TEST_CASE(subscriber_wrong_syntax)
...
+   conn.async_run(cfg, [&](error_code ec) {
+      BOOST_TEST(ec == net::error::operation_aborted);
+      run_finished = true;
+   });
+
+   conn.async_exec(req, resp, [&](error_code ec, std::size_t) {
+      BOOST_TEST(ec == error_code());
+      exec_finished = true;
+      conn.cancel();
+   });
```
mzimbres, line 319: Early feedback: when `async_exec` fails to execute the commands of a request it should complete with an error. I think all adapters work like this.

anarthal: How the hell do tests pass then? lol

anarthal: Hm, none of general_aggregate nor general_simple seem to be setting the output error code on failure.

anarthal:
What should we do with this @mzimbres? It looks like adapters don't set the output error code when they receive an individual error. So `async_exec` succeeds, but the obtained response is an error. This kind of makes sense, since there was no networking problem, but the server answered with an error.

PR #291, opened 2025-07-27
Removes handshaker in favor of asio::deferred

### LINKED ISSUE #290: A response error to HELLO may cause subsequent handshake attempts to fail
If I'm reading [these lines](https://github.com/boostorg/redis/blob/88d8f3c0ca95701eb1f07e2ef11e808b84ce7251/include/boost/redis/detail/resp3_handshaker.hpp#L86) correctly, an error response to a HELLO (maybe due to a user that hasn't been created by a deployment pipeline yet) causes subsequent HELLO attempts to always error, since the error is never cleared.

### LINKED ISSUE #297: HELLO failures are not correctly logged
Since logging happens [here](https://github.com/boostorg/redis/blob/da2f0101d01cf45b848a11a8936acf1d47cc5b4e/include/boost/redis/detail/resp3_handshaker.hpp#L48), if `has_error_in_response()` returns true, we log a success when we have a failure. This is what happens when you specify an incorrect password.

### LINKED ISSUE #298: HELLO commands with a non-default username and an empty password don't trigger authentication
Currently, we don't send the AUTH part of the HELLO command if the password is empty. Using ACLs, it's possible to have a user with a custom username and an empty password (although probably unlikely). I'm adding more tests around this, so I can safely implement this.

anarthal:
* Refactors the handshake process to use asio::deferred instead of a custom composed operation.
* Fixes logging on HELLO error (close #297)
* Fixes a potential problem on reconnection after a HELLO error is encountered (close #290)
* Fixes a race condition in the health checker that could cause it to never exit on cancellation
* Adds support for users with a username different than "default" and an empty password (close #298)
* Adds integration testing for authentication
* Adds unit testing for the hello utility functions


anarthal:
This is still early work, but the idea is getting rid of the entire handshaker class, since the same functionality can be implemented in terms of `asio::deferred`. I think I've found a bug (see #290) while looking at the code.

anarthal:
I've fixed all the small bits I've found around HELLO when doing the refactor.

PR #292, opened 2025-07-27
Implements the writer as an FSM and adds tests

anarthal:


anarthal:
Still early work. While writing this I've realized that we are in dire need of solving cancellation in `async_exec`. Once you merge #286, I'll start working on that. It should allow us to move the responsibility of cancellation from the tasks in make_parallel_group to run_op.

PR #293, opened 2025-08-09
Fix use-after-move bugs when calling async_immediate

ashtum:
@anarthal, There might be a similar issue in Boost.Mysql.

### DIFF
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -110,8 +110,11 @@ struct connection_impl {
                   self.reset_cancellation_state(asio::enable_total_cancellation());
                   continue;  // this action does not require yielding
                case exec_action_type::immediate:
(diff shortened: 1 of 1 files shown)


ashtum:
Or maybe this isn’t a bug. From the implementation, it seems the `detail::initiate_immediate` constructor is called before the move occurs:
https://github.com/boostorg/asio/blob/develop/include/boost/asio/immediate.hpp#L94-L95
I mean usages like this are definitely a bug:
```C++
asio::async_immediate(
    self.get_io_executor(),
    asio::append(std::move(self), ec));
```

anarthal:
Yes, the problem is due to the inner call being executed before self.get_io_executor(). This usage should be fine.

ashtum:
That `asio::append(std::move(self), ...)` (and similar utilities) makes for a footgun in composed operations, it can happen in other operations like `async_read_some` too.

PR #296, opened 2025-09-01
Adds Valkey CIs and docs

### LINKED ISSUE #294: [Question] Long-term compatibility with Valkey?
Hi,

Since Redis has forked and the Valkey project is gaining adoption (DigitalOcean, AWS, Google Cloud, etc. supporting it), I was wondering if there are any long-term plans for Boost.Redis to support Valkey as well.

Specifically:

- Is compatibility with Valkey an explicit design goal for the library, or is the focus strictly Redis?
- Are there known differences that could impact Boost.Redis users if they connect to a Valkey instance?
- Would contributions toward Valkey compatibility be welcomed if divergences appear in the future?

It seems like Valkey is building momentum, so it’d be helpful to understand whether Boost.Redis aims to track both ecosystems.

Thanks!

anarthal:
Adds Valkey CIs

Modifies existing builds to use different database flavors and versions:
* Redis 7.4.5
* Redis 8.2.1
* Valkey 8.1.3

Adds docs on support plans for Valkey

close #294

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,7 @@ jobs:
             cxxstd: '17'
             build-type: 'Debug'
             ldflags: ''
+            server: "redis:7.4.5-alpine"
 
           - toolset: gcc-11
             install: g++-11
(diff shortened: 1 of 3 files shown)


mzimbres:
Many thanks. Just as a note Boost.Redis is officially recommended by Redis [here](https://redis.io/docs/latest/develop/clients/ ) so perhaps we should notify Valkey as well so they perhaps start to officially recommend us as well. I believe that would bring us some users.

PR #301, opened 2025-09-06
Replaces tribool by an enum and adds coverage for multiplexer

anarthal:
* In the context of the multiplexer, replaces tribool by consume_result to enhance readability and make values smaller
* Splits the multiplexer tests out of test_low_level_sync_sans_io into a separate test file
* Increases testing coverage for the multiplexer class

This PR entails no functional change.


anarthal:
My intention is fixing cancellation in `async_exec` next. This PR is intended to make sure that I don't break anything while doing it, since it entails changes to it. Also, it helps me get more familiar with the code. It's still in a draft state.

include/boost/redis/detail/multiplexer.hpp
```diff
@@ -31,7 +31,13 @@ class request;
 
 namespace detail {
 
-using tribool = std::optional<bool>;
+// Return type of the multiplexer::consume_next function
+enum class consume_result
```
mzimbres, line 33: This type could possibly be returned directly by the parser consume operation here https://github.com/boostorg/redis/blob/0cf2441ed2b81506f4b6f626d01ea43785ba31a9/include/boost/redis/resp3/parser.hpp#L80. But since this is hidden from the user at the moment we don't have to do it now.

anarthal: I will do it in subsequent PRs, since this is already pretty big.

PR #303, opened 2025-09-11
Adds custom setup requests

### LINKED ISSUE #302: Make command HELLO optional
Please, make command HELLO optional.

Some servers implementing RESP3 don't implement command HELLO.
And it's impossible to use boost.redis with such servers.

But other client libraries and even redis-cli work with these servers just fine.

Examples of such servers: Tarantool DB and Picodata Radix. Both have modules for RESP3.

I think, library should give possibility to use this command, but should not call it itself by default.

anarthal:
Adds config::setup and config::use_setup, to run arbitrary Redis commands on connection establishment
Improves docs for config::{username, password, clientname, database_index}
Splits all connection establishment tests into test_conn_hello

close #302


include/boost/redis/config.hpp
```diff
@@ -7,12 +7,24 @@
...
 
 namespace boost::redis {
+namespace detail {
+
+inline request make_hello_request()
+{
+   request req;
+   req.push("HELLO", "3");
+   return req;
+}
```
mzimbres: By default I prefer to move every implementation to an `.ipp` file.

include/boost/redis/config.hpp
```diff
@@ -95,14 +131,46 @@ struct config {
...
     * maximum size is reached no more memory allocations are made
     * since the buffer is reused.
     */
    std::size_t read_buffer_append_size = 4096;
+
+   /** @brief Enables using a custom requests during connection establishment.
+    * 
+    * If set to true, the @ref setup member will be sent to the server immediately after
+    * connection establishment. Every time a reconnection happens, the setup
+    * request will be executed. This request can be used to perform authentication,
```
mzimbres: I would add  **before any other requests.** at the end of

> Every time a reconnection happens, the setup request will be executed.

include/boost/redis/connection.hpp
```diff
@@ -337,12 +337,23 @@ class run_op {
...
    {
-      return conn_->async_exec(
-         conn_->hello_req_,
-         any_adapter(conn_->hello_resp_),
-         asio::deferred([&conn = *this->conn_](system::error_code hello_ec, std::size_t) {
-            return asio::deferred.values(on_hello(conn, hello_ec));
-         }))(std::forward<CompletionToken>(token));
+      // clang-format off
+      // The hello pipeline might contain commands even when config::use_hello is false
+      // (e.g. AUTH and SELECT).
```
mzimbres: I am a bit confused by this comment. First I think it means `use_setup` and not `use_hello`. Second it seems to mean
> when config::use_hello is **true**

and not
> when config::use_hello is **false**

Also, it implies that in some cases we will be sending both the internal `hello_req_` and the `setup` requests, but when I look at the implementation I see
```cpp
if (cfg.use_setup) {
   // Use the provided request, but setting the priority flag,
   // even if the request contains no HELLO
   // TODO: could we make this better?
   req = std::move(cfg.setup);
   request_access::set_hello_priority(req, true);
}
```
That means, either one or the other is used but they are never merged.

include/boost/redis/detail/hello_utils.hpp
```diff
@@ -13,7 +13,7 @@
 
 namespace boost::redis::detail {
 
-void setup_hello_request(config const& cfg, request& req);
+void setup_hello_request(config& cfg, request& req);  // might move cfg.setup, for efficiency
```
mzimbres: I prefer not to move in cases where it does not matter from the performance point of view. Can we however just drop [hello_req_](https://github.com/boostorg/redis/blob/0cf2441ed2b81506f4b6f626d01ea43785ba31a9/include/boost/redis/connection.hpp#L89) entirely and use `config::setup`  when `use_setup = false`? We make a config copy internally so it does not matter if we change it.

include/boost/redis/impl/hello_utils.ipp
```diff
@@ -4,31 +4,45 @@
...
  */
 
+#include <boost/redis/config.hpp>
 #include <boost/redis/detail/hello_utils.hpp>
+#include <boost/redis/request.hpp>
 
 namespace boost::redis::detail {
 
-void setup_hello_request(config const& cfg, request& req)
+void setup_hello_request(config& cfg, request& req)
```
mzimbres: I think this should be renamed to `set_request_priority`.

mzimbres:
Thanks. We could perhaps also remove [this](https://github.com/boostorg/redis/blob/0cf2441ed2b81506f4b6f626d01ea43785ba31a9/include/boost/redis/request.hpp#L326-L327) that sets priority conditionally only if is a HELLO command. Looks like an unnecessary complication.

include/boost/redis/connection.hpp
anarthal: This was an unintended leftover from my previous implementation. It made no sense. Thanks for the catch.

anarthal:
> Thanks. We could perhaps also remove [this](https://github.com/boostorg/redis/blob/0cf2441ed2b81506f4b6f626d01ea43785ba31a9/include/boost/redis/request.hpp#L326-L327) that sets priority conditionally only if is a HELLO command. Looks like an unnecessary complication.

Removing the check breaks everything because the priority flag defaults to true in the constructor. Changing it to false is technically a breaking change. In the next PR I will deprecate the priority functions and remove all this in a subsequent release.

anarthal:
> We already have O(n) from the std::stable_partition so sorting would only add an extra Olog(n)) to the complexity. But I agree there isn't any strong usecase for it so far.

I don't know to which extent these O(n) and O(nlogn) are comparable. `stable_partition` is called in:

* `cancel_on_conn_lost`: called once per disconnection. Never during regular operation. We can likely neglect this cost.
* `cancel_waiting`: similar argument to the above.
* `release_push_requests`: called once per write. There is no limit to the number of requests a write may have.

On the other hand, sorting price gets paid once per added request, rather than once per write. So there's a factor that this analysis is not considering.

I _think_ the multiplexer can be modified to use linked lists instead of a vector, and that would yield better efficiency. But I'd need to write the code and benchmark it to be sure, so I haven't done anything yet. In any case, I wouldn't pay this price, at least until some user asks for it.

In a

anarthal:
I think I've addressed all your comments. I've also done a general renaming removing "hello" and "handshake" in favor of "setup request".

mzimbres (approved):

mzimbres:
Thanks.

PR #304, opened 2025-09-14
Removes CMakePresets.json

anarthal:


anarthal:
The presets file contains config for quite old toolchains, and creates problems with my vscode environment (I prefer using kits and settings.json. It looks like a developer-specific file that should probably be excluded from the source tree.

mzimbres:
> It looks like a developer-specific file that should probably be excluded from the source tree.

`CMakePresets.json` is not user specific, that's `CMakeUserPresets.json`. I don't see how it can cause problem but I haven't been using this file for some time now so we can drop it.

anarthal:
Apparently vscode strongly prefers CMakePresets.json to kits when it's present. I didn't know that CMakeUserPresets.json was a thing. Since the presets are out of date, I'll merge this. Thanks.

PR #305, opened 2025-09-15
Deprecates request::config::hello_with_priority

anarthal:


include/boost/redis/request.hpp
```diff
@@ -67,11 +67,17 @@ class request {
...
        * possible to send `HELLO` commands and authenticate before other
        * commands are sent.
+       *
+       * @par Deprecated
+       * This field has been superseded by @ref config::setup.
+       * This setup request will always be run first on connection establishment.
+       * Please use it to run any required setup commands.
+       * This field will be removed in subsequent releases.
        */
       bool hello_with_priority = true;
```
mzimbres, line 82: I have to ask again: We are deprecating but it won't be removed, just renamed to e.g. has_priority and we will hide it from users?

anarthal: Yes. My plan is keeping the member (has_priority_) and the request_access members as we have them today. So only hiding it.

mzimbres (approved):

Issue #104, opened 2023-05-20
Make the health checker more flexible


mzimbres:
At the moment if the response to a health-check does not come under the time specified by the user, the connection will timeout. This is not flexible enough to deal with e.g. heavy load where the specified interval cannot be honored because e.g. it is processing too many pushes.

We can offer two options to users

1. Inflexible checks: The connection times out if the specified interval is not honored. (Current implementation)
2. Flexibel checks: Any incoming data, like server pushes are seen as a proof of health (and not only the response to a health-check).

Option 2. covers my own needs, we might not need option 1.

mzimbres:
I can reproduce and will investigate soon. As a workaround, instead of using `boost::redis::connection` you can use an executor other that the default, for example

```cpp
using executor_type = net::io_context::executor_type;
using connection = boost::redis::basic_connection<executor_type>;
```

This might have cascading effect, for example, you might have to redefine awaitable type as `net::awaitable<void, executor_type>` instead of `net::awaitable<void>`.

anarthal:
I'm reading this issue now and I'm convinced that 2 is the way to go. I believe we could integrate health check in the reader, writer and multiplexer completely, without any need of an extra layer. If I have the time, I'll write a PR.

PR #308, opened 2025-09-18
Adds a connection health flag to the multiplexer

### LINKED ISSUE #236: TLS cancel_if_not_connected unexpected behavior
Hi,

I'm using Boost 1.87, Ubuntu 24.10, clang 18.1.8, Redis 7.0.15.

Iḿ trying to identify if my connection to Redis was a success, and in order to do that I was running PING after call async_run. I noticed, though, two different behaviors, and I use or not use TLS. 127.0.0.5 is not a valid Redis server.

Running  with TLS I got an error, cancel doesn't work and I see on terminal that connection and handshake process is still happening. In the end, I just would like to know if connection as a success or not:

```
config cfg;
   
cfg.addr.host = "127.0.0.1";
cfg.addr.port = "6379";
cfg.username = "default";
cfg.password = "mypsw";
cfg.database_index = 5;
cfg.use_ssl = true;

boost::asio::ssl::context ssl_ctx{boost::asio::ssl::context::tlsv12_client};

ssl_ctx.load_verify_file("/etc/certificates/redis_client/ca.pem");
ssl_ctx.use_certificate_file( "/etc/certificates/redis_client/client.pem", boost::asio::ssl::context::file_format::pem);
ssl_ctx.use_private_key_file( "/etc/certificates/redis_client/client.key", boost::asio::ssl::context::file_format::pem);

request req;
req.get_config().cancel_if_not_connected = true;
req.push("PING", "Hello world");

response<std::string> resp;

asio::io_context ioc;
connection conn{ioc, std::move(ssl_ctx)};

conn.async_run(cfg, {}, [](boost::system::error_code const& ec)
{
    std::cout << "async_run: " << ec.message() << std::endl;
});

conn.async_exec(req, resp, [&](auto ec, auto) {
    if (!ec)
        std::cout << "PING: " << std::get<0>(resp).value() << std::endl;
        std::cout << "CONNECTED" << std::endl;
    else
        std::cout << "Exec error: " << ec.message() << std::endl;
    conn.cancel();
});

ioc.run();
```

Exec error: Not connected.
(Boost.Redis) run-all-op: resolve addresses 127.0.0.1:6379
(Boost.Redis) run-all-op: connected to endpoint 127.0.0.1:6379
(Boost.Redis) Runner: SSL handshake Success
(Boost.Redis) writer-op: 143 bytes written.
(Boost.Redis) reader-op: 170 bytes read.
(Boost.Redis) hello-op: Success
(Boost.Redis) writer-op: 32 bytes written.
(Boost.Redis) reader-op: 18 bytes read.
(Boost.Redis) writer-op: 32 bytes written.
(Boost.Redis) reader-op: 18 bytes read.
(Boost.Redis) writer-op: 32 bytes written.
(Boost.Redis) reader-op: 18 bytes read.
(Boost.Redis) writer-op: 32 bytes written.
(Boost.Redis) reader-op: 18 bytes

Thanks,

Rodrigo.

anarthal:
Uses this flag in async_exec to make cancel_if_not_connected reliable
Deprecates basic_connection::run_is_canceled()
Introduces preconditions in multiplexer methods to be called by the reader and writer tasks
Ensures that request cancellation due to a connection lost is only called after the reader and writer have finished, avoiding potential race conditions

close #236

### DIFF
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -102,7 +102,7 @@ struct connection_impl {
       {
          while (true) {
             // Invoke the state machine
-            auto act = fsm_.resume(obj_->is_open(), self.get_cancellation_state().cancelled());
+            auto act = fsm_.resume(self.get_cancellation_state().cancelled());
 
             // Do what the FSM said
             switch (act.type()) {
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -861,8 +861,8 @@ public:
     */
    void cancel(operation op = operation::all) { impl_->cancel(op); }
 
-   // TODO: handle this
-   auto run_is_canceled() const noexcept { return impl_->mpx_.get_cancel_run_state(); }
+   BOOST_DEPRECATED("This function is internal and will be removed in the next release.")
+   bool run_is_canceled() const noexcept { return !impl_->mpx_.is_connection_healthy(); }
 
    /// Returns true if the connection will try to reconnect if an error is encountered.
    bool will_reconnect() const noexcept { return impl_->will_reconnect(); }
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -174,7 +174,6 @@ struct connection_impl {
       stream_.close();
       writer_timer_.cancel();
       receive_channel_.cancel();
-      mpx_.cancel_on_conn_lost();
    }
 
    bool is_open() const noexcept { return stream_.is_open(); }
@@ -432,9 +431,9 @@ public:
 
             // If we were successful, run all the connection tasks
             if (!ec) {
-               conn_->read_buffer_.clear();
-               conn_->mpx_.reset();
                clear_response(conn_->setup_resp_);
+               conn_->read_buffer_.clear();
+               conn_->mpx_.on_connection_up();
 
                // Note: Order is important here because the writer might
                // trigger an async_write before the setup request is sent,
@@ -461,6 +460,8 @@ public:
 
                // The parallel group result will be translated into a single error
                // code by the specialized operator() overload
+
+               conn_->mpx_.on_connection_down();
 
                // The receive operation must be cancelled because channel
                // subscription does not survive a reconnection but requires
@@ -860,6 +861,7 @@ public:
     */
    void cancel(operation op = operation::all) { impl_->cancel(op); }
 
+   // TODO: handle this
    auto run_is_canceled() const noexcept { return impl_->mpx_.get_cancel_run_state(); }
 
    /// Returns true if the connection will try to reconnect if an error is encountered.
--- a/include/boost/redis/detail/multiplexer.hpp
+++ b/include/boost/redis/detail/multiplexer.hpp
@@ -116,8 +116,27 @@ public:
       std::size_t read_size_;
    };
 
+   // This constructor makes testing easier
+   explicit multiplexer(bool conn_healthy = false);
+
+   // Handle connection health
+   void on_connection_up();  // Might be called in any state
+
+   void on_connection_down();  // Must be called once, with is_connection_healthy() == true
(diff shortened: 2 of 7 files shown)


include/boost/redis/connection.hpp
```diff
@@ -174,7 +174,6 @@ struct connection_impl {
       stream_.close();
       writer_timer_.cancel();
       receive_channel_.cancel();
-      mpx_.cancel_on_conn_lost();
    }
 
    bool is_open() const noexcept { return stream_.is_open(); }
```
mzimbres, line 179: Can we deprecate `is_open()` as well? Or at least is should be renamed to `is_healthy()`.

include/boost/redis/connection.hpp
```diff
@@ -432,9 +431,9 @@ class run_op {
 
             // If we were successful, run all the connection tasks
             if (!ec) {
-               conn_->read_buffer_.clear();
-               conn_->mpx_.reset();
                clear_response(conn_->setup_resp_);
+               conn_->read_buffer_.clear();
+               conn_->mpx_.on_connection_up();
```
mzimbres, line 436: I would like to move away the implementation towards https://github.com/boostorg/redis/issues/104. For this PR that would mean `mpx.on_connection_up()` should be actually be called `mpx.log_data_received()` and be called somehow in `redis_stream::async_read_some` when the read size is different from 0. All my other comments in this PR assume this model.

Edit: perhaps it makes more sense to move health funktionality to the redis_sream where data is actually received.

include/boost/redis/detail/multiplexer.hpp
```diff
@@ -116,8 +116,27 @@ class multiplexer {
       std::size_t read_size_;
    };
 
+   // This constructor makes testing easier
+   explicit multiplexer(bool conn_healthy = false);
+
+   // Handle connection health
+   void on_connection_up();  // Might be called in any state
+
+   void on_connection_down();  // Must be called once, with is_connection_healthy() == true
```
mzimbres, line 125: This function is not needed since the connection is considered down when data is older than the ping interval specified in `config::health_check_interval`.

include/boost/redis/detail/multiplexer.hpp
```diff
@@ -184,9 +187,9 @@ class multiplexer {
    std::deque<std::shared_ptr<elem>> reqs_;
    resp3::parser parser_{};
    bool on_push_ = false;
-   bool cancel_run_called_ = false;
    usage usage_;
    any_adapter receive_adapter_;
+   bool conn_healthy_ = false;
```
mzimbres, line 192: I think this could be a `std::chrono::time_point` that gets update at each `log_data_received()`. If creating a timestamp is two expensive i.e. I would need to think more.

include/boost/redis/detail/multiplexer.hpp
```diff
@@ -116,8 +116,27 @@ class multiplexer {
...
 
+   // This constructor makes testing easier
+   explicit multiplexer(bool conn_healthy = false);
+
+   // Handle connection health
+   void on_connection_up();  // Might be called in any state
+
+   void on_connection_down();  // Must be called once, with is_connection_healthy() == true
+
+   bool is_connection_healthy() const { return conn_healthy_; }
```
mzimbres, line 127: This function should check if conn_healthy_ is not older than `config::health_check_interval`.

anarthal:
I've clearly tried to get too many things together in a single PR :) I'm going to split this in two: getting rid of the race condition with cancel_on_conn_lost and then improve the health checker. I need to think of your comments about the latter.

include/boost/redis/connection.hpp
anarthal: This would be equivalent to calling `redis_stream::async_read_some` with `asio::cancel_after(ping_interval)`, and tearing down the connection if the timeout elapses, wouldn't it?

PR #309, opened 2025-09-19
Fixes a race condition when cancelling requests on connection lost

anarthal:
* Changes how cancel_on_conn_lost is used to ensure it is called only once and after the reader and writer tasks have exited.
* This fixes a problem in test_conn_reconnect
* Adds a test for multiplexer::reset()
* Adds stronger invariants to the multiplexer functions to be called by the reader and writer
* Removes test_issue_181, since the same functionality is being covered by unit tests already
* Removes basic_connection::run_is_canceled


include/boost/redis/impl/multiplexer.ipp
```diff
@@ -196,13 +203,12 @@ std::size_t multiplexer::cancel_waiting()
...
 {
-   // Protects the code below from being called more than
-   // once, see https://github.com/boostorg/redis/issues/181
-   if (std::exchange(cancel_run_called_, true)) {
-      return 0;
-   }
+   // Should only be called once per reconnection.
+   // See https://github.com/boostorg/redis/issues/181
+   BOOST_ASSERT(!cancel_run_called_);
+   cancel_run_called_ = true;
```
mzimbres, line 211: This flag is now used only in asserts? Can we protect its declaration around `#if DEBUG ... #endif`?

include/boost/redis/detail/multiplexer.hpp
```diff
@@ -116,8 +116,10 @@ class multiplexer {
       std::size_t read_size_;
    };
 
-   auto remove(std::shared_ptr<elem> const& ptr) -> bool;
-
+   //
+   // To be called by the reader and the writer. Might only be called
+   // before the connection is torn down (before cancel_on_conn_lost() is called).
+   //
```
mzimbres: This comment looks misplaced. I don't get why prepare_write can only be called by the reader and writer and only before the connection is torn down. It is called on every write operation by the writer.

include/boost/redis/impl/multiplexer.ipp
anarthal: I'm not too fan of changing class layout like this, as IIRC could cause trouble (e.g. when mixing translation units with different compile flags). I think we will get rid of this flag once we review the health checker, and it's still one byte per connection, so I'd say leave as is for now.

include/boost/redis/detail/multiplexer.hpp
anarthal: It's necessary for `cancel_on_conn_lost()` to establish its invariants. The idea is that, after `cancel_on_conn_lost()` is called, all requests are either cancelled or return to the waiting state, so they can be re-sent afterwards. If, for whatever reason, a race condition happens and cancel_on_conn_lost() runs first, and then a `prepare_write()` happens (because the writer hasn't exited yet), then there will be leftover requests in `staged` state. When the connection is re-established, these won't be sent to the server again (since they are `staged`), but there won't be any response coming (since the write happened in the previous connection), causing trouble.

By making sure that all operations that might "progress" a request happen before `cancel_on_conn_lost()` is called, we can make sure that `cancel_on_conn_lost()` leaves everything in a well-defined state.

That said, the comment should probably express what I've written here, rather than what it says today. I'll update it.

PR #310, opened 2025-09-20
Adds reliable cancellation support to async_exec

anarthal:
* Terminal cancellation in async_exec no longer tears down the connection when the cancelled request has been sent to the server.
* Adds support for partial cancellation in async_exec with the same semantics.

### DIFF
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -118,9 +118,6 @@ struct connection_impl {
                case exec_action_type::wait_for_response:
                   notifier_->async_receive(std::move(self));
                   return;
-               case exec_action_type::cancel_run:
-                  obj_->cancel(operation::run);
-                  continue;  // this action does not require yielding
                case exec_action_type::done:
                   notifier_.reset();
                   self.complete(act.error(), act.bytes_read());
--- a/include/boost/redis/detail/multiplexer.hpp
+++ b/include/boost/redis/detail/multiplexer.hpp
@@ -95,6 +95,16 @@ public:
 
       auto get_adapter() -> any_adapter& { return adapter_; }
 
+      // Marks the element as an abandoned request. An abandoned request
+      // won't cause problems when its response arrives, but that response will be ignored.
+      void mark_as_abandoned();
+
+      [[nodiscard]]
+      bool is_abandoned() const
+      {
+         return !req_;
+      }
+
    private:
       enum class status
       {
@@ -116,7 +126,7 @@ public:
       std::size_t read_size_;
    };
 
-   auto remove(std::shared_ptr<elem> const& ptr) -> bool;
+   void cancel(std::shared_ptr<elem> const& ptr);
 
    [[nodiscard]]
    auto prepare_write() -> std::size_t;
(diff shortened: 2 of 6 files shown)


anarthal:
This is a very early prototype, but the idea is changing how async_exec deals with cancellation so it's useful to the users. Right now, cancellation can either break the connection or do nothing, which makes it almost useless. With this done, we can freely use `asio::cancel_after` and go.

The idea is that when a request gets cancelled but it has already been sent to the server, we keep an "abandoned request" item in the multiplexer to signal that we expect a response. When cancellation happens, the user operation finishes immediately, so the request and response objects they used can be freely destroyed. We keep this internal marker so we know that some RESP3 nodes will be incoming. Once they arrive, we discard them and continue with our lives.

I want to merge #309 before this, since it makes reasoning about the multiplexer easier, and reduces the number of tests I need to write.

include/boost/redis/detail/multiplexer.hpp
```diff
@@ -95,6 +95,16 @@ class multiplexer {
 
       auto get_adapter() -> any_adapter& { return adapter_; }
 
+      // Marks the element as an abandoned request. An abandoned request
+      // won't cause problems when its response arrives, but that response will be ignored.
+      void mark_as_abandoned();
```
mzimbres: I think this should follow suit an be named `mark_abandoned`.

include/boost/redis/connection.hpp
```diff
@@ -808,41 +812,48 @@ class basic_connection {
...
+    *
+    *   @li If the request hasn't been sent to the server yet, cancellation will prevent it
+    *       from being sent to the server. In this situation, all cancellation types are supported
+    *       (`asio::cancellation_type_t::terminal`, `asio::cancellation_type_t::partial` and
+    *       `asio::cancellation_type_t::total`).
+    *   @li If the request has been sent to the server but the response hasn't arrived yet,
+    *       cancellation will cause `async_exec` to complete immediately. When the response
+    *       arrives from the server, it will be ignored. In this situation, only
+    *       `asio::cancellation_type_t::terminal` and `asio::cancellation_type_t::partial`
+    *       are supported.
```
mzimbres: Perhaps complement

> In this situation, only `asio::cancellation_type_t::terminal` and `asio::cancellation_type_t::partial`are supported, **`total` cancelation will be ignored**.

mzimbres:
This is such a nice feature that I am wondering if we should add a section in the docs. It could be something along the lines of

> async_exec operation can be cancelled, for example
> 
>    conn.async_exec(req, resp, cancel_after(3s))
> 
> will cause the request to be cancelled if not response was received after 3s have elapsed. If cancellation occurs before the request was sent, it will be remove from the connection internal queue with no side effects. Otherwise if the request was already sent, cancellation is still honored but the response to the cancelled request will be ignored by the connection. Notice that in the latter case the request might have had a side effect in the Redis server. For more details please refer to the async_exec docs.

anarthal:
Completely agree with the docs section :) Except that cancel_after doesn't work yet ^^ we need to add some support for it in initiations, which I'll do in my next PR. I had thought of a broader section "Reliable connections and cancellation" where we could also explain cancel_if_unresponded and friends.

PR #311, opened 2025-09-21
Simplifies the read_buffer and add rotated bytes to usage.

mzimbres:
Data rotation in the read buffer creates latency, we know it is preset but so far its magnitude was unknown. This PR adds it as a new field to `usage`. For example, the `test_conn_echo_stress` outputs now

```
   Commands sent: 780,002
   Bytes sent: 32,670,085
   Responses received: 780,001
   Pushes received: 750,001
   Bytes received (response): 3,210,147
   Bytes received (push): 32,250,036
   Bytes rotated: 3,109,190,184
```

In total approximately 34Mb are received but 3Gb are rotated.


mzimbres:
@anarthal, the amount of data rotated looks completely unacceptable to me and is probably a big source of latency. Once we merge this I will open a ticket to address that problem in the `read_buffer` by introducing a buffer offset that trades data rotation with memory consumption i.e. `read_buffer_.consume()` will update an offset until a threshold is reached and only then rotate.

anarthal:
Hm, we're rotating unconditionally aren't we? Even if there are N messages cached. For a large N, that's almost an O(N2) rotation in terms of data size.

In MySQL I only rotate when the buffer needs extra space, and not after every message is consumed.

include/boost/redis/impl/multiplexer.ipp
```diff
@@ -126,27 +137,48 @@ consume_result multiplexer::consume_next_impl(std::string_view data, system::err
...
-      return std::make_pair(ret, consumed);
+      auto const res = read_buffer_.consume(consumed);
+      commit_usage(ret == consume_result::got_push, res);
+      return std::make_pair(ret, res.consumed);
    }
 
    return std::make_pair(consume_result::needs_more, consumed);
 }
 
+auto multiplexer::prepare_read() noexcept -> system::error_code
```
anarthal, line 175: If these are already defined in an .ipp in read_buffer, it might be more lightweight to define them inline here.

include/boost/redis/usage.hpp
```diff
@@ -36,6 +36,9 @@ struct usage {
 
    /// Number of push-bytes received.
    std::size_t push_bytes_received = 0;
+
+   /// Number of bytes rotated in the read buffer.
+   std::size_t bytes_rotated = 0;
```
anarthal, line 41: Is this something interesting for the user, or just for us?

test/common.hpp
```diff
@@ -36,4 +36,24 @@ void run(
...
+
+// Read data into the multiplexer with the following steps
+//
+//   1. prepare_read
+//   2. get_read_buffer
+//   3. Copy data in the buffer from 2.
+//   4. commit_read;
+//
+// This is used in the multiplexer tests.
+void read(multiplexer& mpx, std::string_view data);
```
anarthal: Could we split these functions into a separate header? This is including asio/awaitable, which is super heavy.

test/common.cpp
```diff
@@ -69,9 +69,22 @@ void run_coroutine_test(net::awaitable<void> op, std::chrono::steady_clock::dura
...
+{
+   auto const ec = mpx.prepare_read();
+   BOOST_ASSERT(ec == system::error_code{});
+   auto const buffer = mpx.get_prepared_read_buffer();
+   BOOST_ASSERT(buffer.size() >= data.size());
+   std::copy(data.cbegin(), data.cend(), buffer.begin());
+   mpx.commit_read(data.size());
+}
+
+void read2(multiplexer& mpx, std::string_view data)
```
anarthal: Is this only used in the reader tests? If so, I'd move it there. Also, read2 is not very semantic - how about making it `copy_to(span<char> to, string_view from)`?

anarthal (commented):
Most of the comments are small observations. My only real question is whether we want to support `usage::rotated_bytes` long-term, or if this is something used internally for benchmarks, only. I'd lean towards the second option. If that's the case, we'd need to clearly state that the member is not for public use, not include it in the reference, or otherwise place it somewhere else, rather than in usage.

include/boost/redis/usage.hpp
mzimbres: The `usage` struct is not useful for the average user, only for those users profiling their app. I don't think they will ever come asking _hey, how much data do you rotate internally_ :) but having the field there make them at least aware there can be some latency hidden there.  At the moment it is entirely meant for us, we will want to compare numbers when we optimize (could actually sax fix) the read buffer.

anarthal: Well, if for whatever reason we want to remove it later, we can leave it as zero and go. So I see no real problem.

PR #316, opened 2025-09-23
Reduces the time elapsed by test_conn_health_check and adds test_conn_monitor

anarthal:
Splits test_conn_health_check into a test to verify that the health checker works and another to verify that MONITOR is handled correctly.
Tests no longer wait or use CLIENT PAUSE


test/test_conn_check_health.cpp
```diff
@@ -5,133 +5,81 @@
...
-using boost::redis::operation;
-using boost::redis::generic_response;
-using boost::redis::consume_one;
 using namespace std::chrono_literals;
 
-// TODO: Test cancel(health_check)
-
 namespace {
 
-struct push_callback {
```
mzimbres, line 34: Glad this is being removed, it is a mess.

test/test_conn_monitor.cpp
```diff
@@ -0,0 +1,121 @@
...
+using boost::redis::operation;
+using boost::redis::generic_response;
+using boost::redis::consume_one;
+using namespace std::chrono_literals;
+
+namespace {
+
+// Verifies that using the MONITOR command works properly.
+// Opens a connection, issues a MONITOR, issues some commands to
+// generate some traffic, and waits for several MONITOR messages to arrive.
```
mzimbres, line 33: > // Opens a connection, issues a MONITOR, issues some commands to
> // generate some traffic, and waits for several MONITOR messages to arrive.

I thought it was not allowed to issue commands on the same connection one MONITOR mode is entered. Perhaps this was only a restriction in RESP2 since it does not support server pushes.

mzimbres (approved):

PR #317, opened 2025-09-26
Simplifies the health checker

anarthal:
Modifies the health checker to use asio::cancel_after rather than a separate parallel group task


include/boost/redis/connection.hpp
```diff
@@ -298,6 +300,85 @@ struct reader_op {
...
+
+         for (;;) {
+            // Clean up any previous leftover
+            clear_response(conn_->ping_resp_);
+
+            // Execute the request
+            BOOST_ASIO_CORO_YIELD
+            {
+               auto* conn = conn_;  // avoid use-after-move problems
+               auto timeout = 2 * conn->cfg_.health_check_interval;
```
mzimbres: The `2 *` factor should be removed now to [match the docs](https://github.com/boostorg/redis/blob/f955dc01d2c5b9695f347992c0136321a8ff5276/include/boost/redis/config.hpp#L107-L109).

include/boost/redis/connection.hpp
```diff
@@ -298,6 +300,85 @@ struct reader_op {
...
+            if (is_cancelled(self)) {
+               conn_->logger_.trace("ping_op (2): cancelled");
+               self.complete(asio::error::operation_aborted);
+               return;
+            }
+
+            // Check for errors in PING
+            // TODO: transforming operation_aborted into pong_timeout
+            // is unreliable until we migrate everything to use per-operation cancellation.
+            ec = check_ping_response(ec, conn_->ping_resp_);
```
mzimbres: I think `ec` should not be an argument to `check_ping_response`. It should be treated explicitly above and convert an operation aborted into `pong_timeout` e.g.

```cpp
if (ec) {
  assert(ec == operation_canceled); // I guess there should be no other type of errors here?
  self.complete(pong_timeout);
}

ec = check_ping_response(conn_->ping_resp_);
...
```

include/boost/redis/connection.hpp
```diff
@@ -380,23 +470,22 @@ class run_op {
...
       Self& self,
       order_t order,
-      system::error_code ec0,
-      system::error_code ec1,
-      system::error_code ec2,
-      system::error_code ec3,
+      system::error_code setup_ec,
+      system::error_code health_check_ec,
+      system::error_code reader_ec,
       system::error_code)
```
mzimbres: I believe the error omitted here is the `writer_ec`, we should perhaps not omit it for better readability If necessary we can ignore it with `boost::core::ignore_unused`. I am however wondering why it is not handled below.

anarthal: This will be addressed in my next PR on cancellation.

include/boost/redis/impl/ping_request_utils.ipp
```diff
@@ -0,0 +1,33 @@
...
+system::error_code check_ping_response(system::error_code io_ec, const generic_response& resp)
+{
+   if (io_ec)
+      return io_ec;
+
+   // TODO: we should probably introduce a better error code/logging
+   // for PONG errors. We defer this until
+   // https://github.com/boostorg/redis/issues/104
+   if (resp.has_error())
+      return error::pong_timeout;
```
mzimbres: AFAICS if `resp` has an error it will be a RESP3 error which should not be converted into a `pong_timeout`. The latter should occur only if `cancel_after` fires. An error in this `resp` object should be returned to the user when the `headl_check` operation completes and its [diagnostic](https://github.com/boostorg/redis/blob/f955dc01d2c5b9695f347992c0136321a8ff5276/include/boost/redis/adapter/result.hpp#L26) should be logged.

include/boost/redis/connection.hpp
anarthal: Not that I can think of, but it's I/O and system calls, so you never know. Best practice says to handle unknown errors, too.

mzimbres: Do you mean something like this?
```cpp
if (ec == operation_canceled) {
  self.complete(pong_timeout);
} else {
  self.complete(ec);
}
```
Correction

```cpp
if (ec == operation_canceled) {
  self.complete(pong_timeout);
} else if (!!ec) {
  self.complete(ec);
}
```

anarthal: Yes. I'm refactoring it to a (private) function for sanity.

anarthal:
I think I've addressed all of your comments.

mzimbres (approved):

Issue #319, opened 2025-09-27
Support per-operation cancellation in async_run


anarthal:
This would enable fixing #318 and would make us support `conn.async_run(cfg, asio::cancel_after(10s))` for free.

PR #320, opened 2025-09-27
Implements connect as a FSM and fixes cancellation

anarthal:
Implements redis_stream::async_connect as a FSM
Adds per-operation cancellation handling code
Adds tests

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -17,9 +17,7 @@ endmacro()
 
 macro(make_testable_example EXAMPLE_NAME STANDARD)
   make_example(${EXAMPLE_NAME} ${STANDARD})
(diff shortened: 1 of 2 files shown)


anarthal:
This is an early prototype yet. I've started implementing #319 and it was getting too big and messy around `redis_stream::async_connect`. This implements the required code to handle per-operation cancellation correctly in `async_connect`, while covering this code with tests.

anarthal:
I think this one is actually good - it's just ultra segmented in terms of commits, but they're all part of the PR. I always do squash merging when merging, so these won't appear in history - just a big commit replacing them.

I did something wrong in the cancel_after PR which I think I've solved now.

anarthal:
(looks like I commited once per test and wrote about 20 tests, so... :))

mzimbres (approved):

PR #321, opened 2025-10-03
Adds per-operation cancellation support to async_run and cleans up cancellation

### LINKED ISSUE #318: Race conditions in connection::cancel()
The cancellation mechanism in `connection::cancel()` is currently quite rudimentary, and subject to race conditions. If cancellation is called after an operation has completed, but before its handler is called, `connection::cancel()` does nothing. The correct way to handle this is using per-operation cancel.

### LINKED ISSUE #319: Support per-operation cancellation in async_run

anarthal:
* Adds support for terminal and partial cancellation to async_run.
* Makes basic_connection::cancel use per-operation cancellation under the hood.
* Fixes a number of race conditions during cancellation which could cause the cancellation to be ignored. This could happen if the cancellation is delivered while an async handler is pending execution.
* Deprecates operation::{resolve, connect, ssl_handshake, reconnection, health_check}, in favor of operation::run. Calling basic_connection::cancel with these values (excepting reconnection) is now equivalent to using operation::run.
* Fixes a problem in the health checker that caused ping timeouts to be reported as cancellations.
* Sanitizes how the parallel group computes its final error code.
* Simplifies the reader, writer and health checker to not care about connection cancellation. This is now the responsibility of the parallel group.
* Removes an unnecessary setup_cancellation action in the reader FSM.
* Adds documentation regarding per-operation cancellation to async_receive.
* Adds additional health checker tests.
* Adds async_run per-operation cancellation tests.
* Adds reader FSM cancellation tests.
* Makes test_conn_exec_retry tests more resilient.
* Removes leftovers in the UNIX and TLS reconnection tests. These were required due to race conditions that have already been fixed.

close #318 
close #319


anarthal:
This is a big rework on how cancellation is handled in the connection. It has the side effect of implementing per-operation cancellation for async_run, but the main point is fixing race conditions and making the reader/writer not concerned about cancelling the connection.

I prefer merging this before #320, which fixes the last race conditions regarding cancellation. Once this is done, I'll submit a PR implementing support for cancel_after in connection.

include/boost/redis/connection.hpp
```diff
@@ -339,8 +352,11 @@ struct health_checker_op {
       {
          if (conn_->cfg_.health_check_interval == std::chrono::seconds::zero()) {
             conn_->logger_.trace("ping_op (1): timeout disabled.");
-            BOOST_ASIO_CORO_YIELD asio::async_immediate(self.get_io_executor(), std::move(self));
-            self.complete(system::error_code{});
+
+            // Wait until we're cancelled. This simplifies parallel group handling a lot
+            conn_->ping_timer_.expires_at((std::chrono::steady_clock::time_point::max)());
```
mzimbres, line 357: This is a good idea. But I guess it would have been even better to disable health-checks by letting users set `max()` instead of `zero()`, then we would not need this branch at all. We would only have to reformulate the loop below to first wait and then to ping and not the other way around. That would be a breaking change hard to get around. We would have to rename `health_check_interval` to something like `health_check_interval2` to intentionally break user code.

include/boost/redis/connection.hpp
```diff
@@ -412,24 +416,46 @@ inline system::error_code check_config(const config& cfg)
    return system::error_code{};
 }
 
+inline system::error_code translate_parallel_group_errors(
```
mzimbres: I know `inline` functions might result in better performance but for this specific case it won't make a difference. I want to make compilation as fast as possible, so I prefer to move this to an `.ipp` file.

include/boost/redis/connection.hpp
anarthal: Hm, that wouldn't work because the interval is a duration, not a time_point. I actually had a bug at some point where I used `duration::max` rather than `time_point::max` and it caused an overflow error, with the operation finishing immediately.

include/boost/redis/connection.hpp
```diff
@@ -730,31 +777,52 @@ class basic_connection {
...
       impl_->mpx_.set_config(cfg);
 
-      return asio::async_compose<CompletionToken, void(system::error_code)>(
+      // If the token's slot has cancellation enabled, it should just emit
+      // the cancellation signal in our connection. This lets us unify the cancel()
+      // function and per-operation cancellation
+      auto slot = asio::get_associated_cancellation_slot(token);
+      if (slot.is_connected()) {
+         slot.template emplace<detail::run_cancel_handler<Executor>>(*impl_);
+      }
```
mzimbres, line 801: IIUC this code requires the user slot to not outlive the connection i.e. users should not emit a signal on it if the connection ceased to exist. I am not sure but some people might think this would be a noop instead of crash, so perhaps a note on the docs?

include/boost/redis/connection.hpp
```diff
@@ -80,6 +85,7 @@ struct connection_impl {
    timer_type reconnect_timer_;  // to wait the reconnection period
    timer_type ping_timer_;       // to wait between pings
    receive_channel_type receive_channel_;
+   asio::cancellation_signal run_signal_;
```
mzimbres, line 88: Would this still be needed if we remove the cancel function a let would work entirely with user provided `cancellation_signals`?

mzimbres:
Some many nice things. The simplifications on the reader_fsm are relieving. Good job, thanks.

include/boost/redis/connection.hpp
anarthal: Actually, this is what happens with all Asio objects. I also thought that it should be a noop (and went down a very hard path to achieve this in MySQL). And yet a single timer doesn't handle this. I've just checked with ASAN, and emitting a cancel signal on a timer after it's been destroyed accesses freed memory.

In any case, I think you're right so I'm gonna document it, at least.

include/boost/redis/connection.hpp
anarthal: No, it wouldn't. It only exists to keep compatibility with `cancel()`. But I feel removing it is breaking too much (essentially, everyone).

PR #322, opened 2025-10-03
Makes the CI run part of the examples

anarthal:
Removes a conditional from CMake that was causing examples to never be run under CI.


anarthal:
This is straightforward: we weren't running any examples in the CI because of an incorrect CMake setting, and this fixes it. The TLS example references your occase.de server, which seems to be down, so I've set it to not testable.

mzimbres (approved):

PR #324, opened 2025-10-03
Adds support for asio::cancel_after

### LINKED ISSUE #226: Support asio::cancel_after
This is a very handy token, but it requires the initiation function object to have a bound executor. The following should work:

```cpp
redis::connection conn (ctx);
co_await conn.async_exec(req, res, asio::cancel_after(10s));
```

anarthal:
* Adds support for asio::cancel_after in connection::{async_exec, async_run}
* Adds cancel_after tests
* Adds an example on using asio::cancel_after
* Adds a discussion page on timeouts and the `cancel_if_unresponded` flag

close #226

### DIFF
--- a/doc/modules/ROOT/pages/cancellation.adoc
+++ b/doc/modules/ROOT/pages/cancellation.adoc
@@ -1,0 +1,55 @@
+//
+// Copyright (c) 2025 Marcelo Zimbres Silva (mzimbres@gmail.com),
+// Ruben Perez Hidalgo (rubenperez038 at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+= Cancellation
+
+Requests may take a very long time. If the server is down, they may suspend forever,
+waiting for the server to be up. Fortunately, requests can be cancelled after
+a certain time using asio::cancel_after:
+
+```
+request req;
+// ...
+
+co_await conn.async_exec(req, resp, asio::cancel_after(10s));
+```
+
+If the request hasn't been responded after 10 seconds, it will
+fail with `asio::error::operation_aborted`. With the coroutine
+usage above, this means a `boost::system::system_error` exception
+with the error code mentioned above.
+
+== Retrying idempotent requests
+
+By default, the library waits until the server is up,
+and then sends the request. But what happens if there is a communication
+error after sending the request, but before receiving a response?
+
+In this situation, we don't know if the request was processed by the server or not.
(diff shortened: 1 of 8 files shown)


doc/modules/ROOT/pages/cancellation.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+== Retrying idempotent requests
+
+By default, the library waits until the server is up,
+and then sends the request. But what happens if there is a communication
+error after sending the request, but before receiving a response?
+
+In this situation, we don't know if the request was processed by the server or not.
+And we have no way to know it. By default, the library mark these requests as
+failed with `asio::error::operation_aborted`. (TODO: do we want another error code here?).
```
mzimbres, line 35: > the library mark these

Needs an **s** at the end.

> do we want another error code here?

I think we do, otherwise the user might not know why it was canceled. Transforming `operation_aborted` into e.g. `unresponded_on_connection_lost` will make it clear.

doc/modules/ROOT/pages/cancellation.adoc
```diff
@@ -0,0 +1,55 @@
...
+You can do so by setting request::config::cancel_if_unresponded to false:
+
+```
+request req;
+req.push("SET", "my_key", 42); // idempotent
+req.get_config().cancel_on_connection_lost = false; // TODO: we shouldn't need this
+req.get_config().cancel_if_unresponded = false; // retry
+
+// Makes sure that the key is set, even in the presence of network errors.
+// This may suspend for an unspecified period of time if the server is down.
```
mzimbres: > This may suspend for an unspecified period of time if the server is down.

I would rephrase this to something like

> This will remain suspended until the server is capable of serving requests again e.g. after a process manager restarted the server.

include/boost/redis/connection.hpp
```diff
@@ -1358,6 +1349,34 @@ class connection {
    }
 
 private:
+   // Function object to initiate the async ops that use asio::any_completion_handler.
+   // Required for asio::cancel_after to work.
```
mzimbres, line 1373: Why is this required? i.e. why wouldn't the lambda be enough?

mzimbres (approved):

include/boost/redis/connection.hpp
anarthal: cancel_after (without a timer argument) requires an executor to construct a temporary timer for the timeout. This executor is taken from the initiation object. This gets done for free if you use async_compose, but needs to be done manually when using async_initiate.

doc/modules/ROOT/pages/cancellation.adoc
anarthal: Opened #326 to track this.

PR #328, opened 2025-10-09
Makes health checks flexible so they don't tear down connections under heavy load

### LINKED ISSUE #104: Make the health checker more flexible

anarthal:
Adds `error::write_timeout`

close #104

### DIFF
--- a/example/cpp20_chat_room.cpp
+++ b/example/cpp20_chat_room.cpp
@@ -10,6 +10,7 @@
 #include <boost/asio/consign.hpp>
 #include <boost/asio/detached.hpp>
 #include <boost/asio/posix/stream_descriptor.hpp>
+#include <boost/asio/read_until.hpp>
 #include <boost/asio/redirect_error.hpp>
 #include <boost/asio/signal_set.hpp>
 
--- a/example/cpp20_echo_server.cpp
+++ b/example/cpp20_echo_server.cpp
@@ -9,6 +9,7 @@
 #include <boost/asio/co_spawn.hpp>
 #include <boost/asio/consign.hpp>
 #include <boost/asio/detached.hpp>
+#include <boost/asio/read_until.hpp>
 #include <boost/asio/redirect_error.hpp>
 #include <boost/asio/signal_set.hpp>
 


include/boost/redis/config.hpp
```diff
@@ -83,7 +83,7 @@ struct config {
     */
    std::optional<int> database_index = 0;
 
-   /// Message used by the health-checker in @ref boost::redis::basic_connection::async_run.
+   /// Message used by `PING` commands sent by the health checker.
    std::string health_check_id = "Boost.Redis";
```
mzimbres, line 87: Sidenote: When the user does not provide one id, I think we should format this to contain the id returned by the `HELLO` command.

include/boost/redis/connection.hpp
```diff
@@ -216,23 +216,40 @@ struct writer_op {
...
+         case writer_action_type::write_some:
+         {
+            auto buf = asio::buffer(conn->st_.mpx.get_write_buffer().substr(act.write_offset()));
+            if (auto timeout = act.timeout(); timeout.count() != 0) {
+               conn->stream_.async_write_some(
+                  buf,
+                  asio::cancel_after(conn->writer_timer_, timeout, std::move(self)));
+            } else {
+               conn->stream_.async_write_some(buf, std::move(self));
+            }
```
mzimbres: Can we unify the two `async_write_some` calls into one by creating an infinite interval when `timeout == 0`?

include/boost/redis/connection.hpp
```diff
@@ -216,23 +216,40 @@ struct writer_op {
...
+            return;
+         }
+         case writer_action_type::wait:
+            if (auto timeout = act.timeout(); timeout.count() != 0) {
+               conn->writer_cv_.expires_after(timeout);
+            } else {
+               conn->writer_cv_.expires_at((std::chrono::steady_clock::time_point::max)());
+            }
```
mzimbres: To avoid this `if - else` here I would move this logic into an action member function and call 
```cpp
conn->writer_cv_.expires_at(act.get_interval())
```

include/boost/redis/config.hpp
anarthal: What does it mean "does not provide one id" here? Leave `health_check_id` to the default? Make `health_check_id` empty?

The change is not trivial because it requires coordination between the setup task (which may not contain a HELLO at all) and the health checker.

include/boost/redis/connection.hpp
anarthal: We can't use a max interval because that causes an integer overflow and UB. `cancel_after` computes a `time_point expiry = now() + timeout`. What we could do is using `cancel_at` and the following logic:

```cpp
asio::cancel_at(timeout.count() == 0 ? (std::chrono::steady_clock::time_point::max)() : std::chrono::steady_clock::now() + timeout)
```

`cancel_at` does cause an allocation, so technically not calling it is lighter than calling it with an infinite time point. I guess calling `async_write_some` with both `cancel_after` and without it causes additional instantiations, increasing build times and executable size. So there's a trade-off here. I can go with the solution I proposed above.

include/boost/redis/connection.hpp
anarthal: `expires_at` requires a `time_point` rather than an interval (duration). We could return a `time_point` in the action, but I'm not too comfortable with it because that requires reading the clock in the FSM, and it complicates testing a lot (there's no easy way to mock reading the clock). What I could do is re-use the approach I proposed above, factoring into a function, to avoid the conditional.

anarthal:
> I am still studying the code but I think the writter-offset should be moved down to the multiplexer where the writer buffer is located. The writer-fsm can then call `mpx.commit_write_some(bytes)`. Then the writer_op wouldn't need this
> 
> ```c++
> auto buf = asio::buffer(conn->st_.mpx.get_write_buffer().substr(act.write_offset()));
> ```
> 
> because the multiplexer knowns about the offset, so `conn->st_.mpx.get_write_buffer()` would already be correct.

I think this is a very good idea. I will work on it.

anarthal:
Comments applied.

include/boost/redis/config.hpp
mzimbres: I mean, if the user does not provide its own `health_check_id` the connection could set it to "Boost.Redis (conn-id)". At the moment we only set to "Boost.Redis".

mzimbres (approved):

include/boost/redis/config.hpp
anarthal: We'd need to detect whether the setup request contains a hello, and then parse the response and update the ping ID. Do you think it's worth it?

PR #333, opened 2025-10-17
Moves the setup request execution to run_fsm

anarthal:
Adds unit tests to cover setup request execution in run_fsm
Entails no functional change


anarthal:
This moves the setup request execution to the sans-io world. It uses an approach similar to the one I used for PINGs in the flexible health-checker.

This

* Simplifies the parallel group handling a lot, because we no longer have a task that finishes before the others.
* Moves more functionality to .ipps. Concretely, after merging this and #328, all logging happens in ipp files. This will allow us to remove the `connection_logger` class completely, separating all logging functionality into the FSM ipp files.

PR #334, opened 2025-10-17
Changes request default's cancel_on_connection_lost to false

anarthal:
Changes request default constructor to set cancel_on_connection_lost to false, matching request::config's default initializer
Overwrites this flag to true for the setup request
Adds unit tests for the latter


anarthal:
This is a fix for #329, where I missed that the request constructor's default argument passes the flag values explicitly. I forgot to update this value, effectively rendering the PR ineffective in most cases. Apparently passing these flags explicitly is required because `config` is a nested type, rather than a standalone type.

I've also made sure to overwrite the flags for the setup request, so it's cancelled on error.

PR #340, opened 2025-10-25
Concludes the work on the generic_flat_response started by Nikolai Vladimirov

mzimbres:
@anarthal, let us start the review for `generic_flat_response`. I need the `async_receive2` so I committed my own implementation and will rebase on top of yours if you find time to implement it.

### DIFF
--- a/include/boost/redis/response.hpp
+++ b/include/boost/redis/response.hpp
@@ -10,6 +10,7 @@
 #include <boost/redis/adapter/result.hpp>
 #include <boost/redis/resp3/node.hpp>
 #include <boost/redis/resp3/tree.hpp>
+#include <boost/redis/resp3/flat_tree.hpp>
 
 #include <boost/system/error_code.hpp>
 
@@ -31,6 +32,9 @@ using response = std::tuple<adapter::result<Ts>...>;
  *  view of the response tree.
  */
 using generic_response = adapter::result<resp3::tree>;
+
+/// Similar to @ref boost::redis::generic_response but stores data contiguously.
+using generic_flat_response = adapter::result<resp3::flat_tree>;
 
 /** @brief (Deprecated) Consume on response from a generic response
  *
--- a/example/cpp20_chat_room.cpp
+++ b/example/cpp20_chat_room.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2022 Marcelo Zimbres Silva (mzimbres@gmail.com)
+/* Copyright (c) 2018-2025 Marcelo Zimbres Silva (mzimbres@gmail.com)
  *
  * Distributed under the Boost Software License, Version 1.0. (See
  * accompanying file LICENSE.txt)
@@ -32,7 +32,7 @@ using boost::asio::dynamic_buffer;
 using boost::asio::redirect_error;
 using boost::redis::config;
 using boost::redis::connection;
-using boost::redis::resp3::flat_tree;
+using boost::redis::generic_flat_response;
 using boost::redis::request;
 using boost::system::error_code;
 using namespace std::chrono_literals;
@@ -45,7 +45,7 @@ auto receiver(std::shared_ptr<connection> conn) -> awaitable<void>
    request req;
    req.push("SUBSCRIBE", "channel");
 
-   flat_tree resp;
+   generic_flat_response resp;
    conn->set_receive_response(resp);
 
    while (conn->will_reconnect()) {
@@ -58,12 +58,12 @@ auto receiver(std::shared_ptr<connection> conn) -> awaitable<void>
          if (ec)
             break;  // Connection lost, break so we can reconnect to channels.
 
-         for (auto const& elem: resp.get_view())
+         for (auto const& elem: resp.value().get_view())
             std::cout << elem.value << "\n";
 
          std::cout << std::endl;
 
-         resp.clear();
+         resp.value().clear();
       }
    }
 }
--- a/example/cpp20_subscriber.cpp
+++ b/example/cpp20_subscriber.cpp
@@ -21,7 +21,7 @@
 namespace asio = boost::asio;
 using namespace std::chrono_literals;
 using boost::redis::request;
-using boost::redis::resp3::flat_tree;
+using boost::redis::generic_flat_response;
 using boost::redis::config;
 using boost::system::error_code;
 using boost::redis::connection;
@@ -49,7 +49,7 @@ auto receiver(std::shared_ptr<connection> conn) -> asio::awaitable<void>
    request req;
    req.push("SUBSCRIBE", "channel");
 
-   flat_tree resp;
+   generic_flat_response resp;
    conn->set_receive_response(resp);
 
    // Loop while reconnection is enabled
@@ -66,12 +66,12 @@ auto receiver(std::shared_ptr<connection> conn) -> asio::awaitable<void>
 
          // The response must be consumed without suspending the
          // coroutine i.e. without the use of async operations.
-         for (auto const& elem: resp.get_view())
+         for (auto const& elem: resp.value().get_view())
             std::cout << elem.value << "\n";
 
          std::cout << std::endl;
 
-         resp.clear();
+         resp.value().clear();
       }
    }
 }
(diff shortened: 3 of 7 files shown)


anarthal:
Do we have any numbers comparing the performance of this approach (converting to string_view immediately after parsing) vs. the custom iterator approach (converting to string_view on iteration)?

I will have a look at this tomorrow

mzimbres:
> Do we have any numbers comparing the performance of this approach (converting to string_view immediately after parsing) vs. the custom iterator approach (converting to string_view on iteration)?

IMO this is not worth the effort. Both cases must contruct construct the `string_view` but the custom iterator approach must copy more data i.e. `node&` vs `node` so it can only be slower.

Performance however is not my main concern because it is probably in sub microsend difference, which is hard to measure.  In essence I regard the `T at(std::size_t);` api to be inferior to `T const& at(std::size_t)` since the latter requires reconstructing the node on each call (imagine needing to call it into a loop). Changing the container is probably never going to be needed, but I at least the current API does not make it impossible.
 
> I will have a look at this tomorrow

Thanks. Please, take into account I hurried a bit to ovoid delaying discussion another week, specially on what we should do with `async_receive2`.

test/CMakeLists.txt
```diff
@@ -4,10 +4,10 @@ add_library(boost_redis_project_options INTERFACE)
 target_link_libraries(boost_redis_project_options INTERFACE boost_redis)
 if (MSVC)
   # C4459: name hides outer scope variable is issued by Asio
-  target_compile_options(boost_redis_project_options INTERFACE /bigobj /W4 /WX /wd4459)
+  target_compile_options(boost_redis_project_options INTERFACE /bigobj /W4 /WX /wd4459 /w34996)
```
anarthal: We shouldn't need this or -Wno-deprecated-declarations. We already define `BOOST_ALLOW_DEPRECATED` in the tests, and examples shouldn't use deprecated features.

include/boost/redis/generic_flat_response_value.hpp
```diff
@@ -0,0 +1,91 @@
...
+#ifndef BOOST_REDIS_FLAT_RESPONSE_HPP
+#define BOOST_REDIS_FLAT_RESPONSE_HPP
+
+#include <boost/redis/adapter/result.hpp>
+#include <boost/redis/resp3/node.hpp>
+
+#include <boost/system/error_code.hpp>
+
+#include <string>
+#include <tuple>
```
anarthal: Unused include

include/boost/redis/resp3/flat_tree.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+   /** @brief Returns the number of times reallocation took place
+    *
+    *  Each call to the `push` member function might result in a
+    *  memory reallocation.  This function returns how many
+    *  reallocations were detected and can be useful to determine how
+    *  much memory to reserve upfront.
+    */
+   auto get_reallocs() const noexcept
+      { return reallocs_; }
```
anarthal, line 84: Better use explicit return type here

include/boost/redis/connection.hpp
```diff
@@ -1028,12 +1093,21 @@ class connection {
 
    /// @copydoc basic_connection::async_receive
    template <class CompletionToken = asio::deferred_t>
+   BOOST_DEPRECATED("Please use async_receive2 instead.")
```
anarthal, line 1107: You probably need `-Wno-deprecated-declarations` and friends because `impl_.async_receive` and `impl_.receive` are deprecated. This means that the user is going to see these warnings when they build `src.hpp`, and we don't want this.

The way I usually solve this is by moving the function from `basic_connection` to `connection_impl`, then calling this second function from both `basic_connection` and `connection`.

include/boost/redis/generic_flat_response_value.hpp
```diff
@@ -0,0 +1,91 @@
...
+    *  Each call to the `push` member function might result in a
+    *  memory reallocation.  This function returns how many
+    *  reallocations were detected and can be useful to determine how
+    *  much memory to reserve upfront.
+    */
+   auto get_reallocs() const noexcept
+      { return reallocs_; }
+
+   // Notify the object that all nodes were pushed.
+   void notify_done();
```
anarthal: These should be made private with a detail accessor. But this is not urgent, I can do it afterwards.

test/test_conn_push.cpp
```diff
@@ -98,10 +99,12 @@ BOOST_AUTO_TEST_CASE(push_received1)
    net::io_context ioc;
    auto conn = std::make_shared<connection>(ioc);
```
anarthal, line 100: We should retain at least one test for the deprecated `async_receive` and `receive` so we make sure we don't break them by accident. I suggest duplicating this test to get a minimum of coverage.

include/boost/redis/response.hpp
```diff
@@ -70,13 +81,11 @@ using generic_response = adapter::result<std::vector<resp3::node>>;
...
  */
+BOOST_DEPRECATED("This function is not needed anymore to consume server pushes.")
 void consume_one(generic_response& r, system::error_code& ec);
 
-/**
- * @brief Throwing overload of `consume_one`.
- *
- * @param r The response to modify.
- */
+/// @copydoc consume_one
```
anarthal: There seems to be a MrDocs bug with copydoc and this seems to fail. It doesn't support Doxygen's overload resolution either (https://github.com/cppalliance/mrdocs/issues/842). Since this is deprecated, I suggest keeping the old docs stating that the function is deprecated and go.

include/boost/redis/response.hpp
```diff
@@ -29,9 +30,19 @@ using response = std::tuple<adapter::result<Ts>...>;
...
-/** @brief Consume on response from a generic response
+/** @brief A generic response that stores data contiguously
+ *  @ingroup high-level-api
+ * 
+ *  Provides the same data format as the generic_response except that
+ *  its strings are views into an internal buffer that stores data
+ *  contiguously, this helps reducing memory allocations and improves
+ *  memory reuse.
+ */
+using generic_flat_response = adapter::result<generic_flat_response_value>;
```
anarthal: New classes should be added to the reference page. If you'd like, please open an issue and I'll add it after you merge it.

include/boost/redis/resp3/flat_tree.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+   // Notify the object that all nodes were pushed.
+   void notify_done();
+
+   /// Returns the number of complete RESP3 messages contained in this object.
+   std::size_t get_total_msgs() const noexcept
+      { return total_msgs_; }
```
anarthal, line 88: I completely forgot - but this class copy semantics are wrong. You will get string_views pointing to the old class. You will need to implement a custom copy constructor and copy assignment. You might also mark these operations as deleted for now and implement them later on.

include/boost/redis/connection.hpp
mzimbres: The problem is that `connection` calls on `basic_connection` functions which call on `connection_impl` functions. So in connection we have
```cpp
std::size_t receive(system::error_code& ec) { return impl_.receive(ec); }
```
and `impl_.receive` is deprecated in `basic_connection`. I will check if I can use `impl.impl.receive()` or something.

anarthal: IIRC you can because they're friend classes (I added that because of a similar problem)

mzimbres:
@D0zee The PR is finally ready for merging. I have kept all your original commits so that you appear as contributor. You are also listed in the copyright notice in the relevant files, please check if everything is fine. I will merge it this week. Thanks.

@anarthal You can also have a final look if you want. I have added a new typedef
```cpp
using tree = std::vector<node>;
```
Which we will have to use in you sentinel PR.

mzimbres:
> Not a big fan of the aliases - I think vector<node> communicates intent better.

For some background. Currently we have the following RESP3 types for a sequence of nodes

  - `std::vector<node>`
  - `flat_tree`
  - `std::vector<node_view>`

IMO these namings do not convey in a user friendly way that these types provide the same data layout. Therefore I aliased 1. and 3. to `tree` and `view_tree` ending up with

  - `tree`
  - `flat_tree`
  - `view_tree`

The generic responses can now be defined more intuitively as

```cpp
using generic_response = result<tree>;
using generic_flat_response = result<flat_tree>;
```

PR #345, opened 2025-10-30
Adds Sentinel support

### LINKED ISSUE #229: cpp20_resolve_with_sentinel.cpp example fails on response parsing
Hello,
This example (cpp20_resolve_with_sentinel.cpp) incorrectly parses Sentinel server responses. Here are my logs:
```
(Boost.Redis) run-all-op: resolve addresses 192.168.88.81:26379
(Boost.Redis) run-all-op: connected to endpoint 192.168.88.81:26379
(Boost.Redis) writer-op: 161 bytes written.
(Boost.Redis) reader-op: 229 bytes read.
(Boost.Redis) reader-op: parse error. Exiting ...
(Boost.Redis) hello-op: Success
(Boost.Redis) writer-op: canceled (3). Exiting ...
(Boost.Redis) run-op: Got RESP3 simple-error. (reader), Success (writer)
(Boost.Redis) ping_op: error/cancelled (2).
(Boost.Redis) check-timeout-op: error/canceled. Exiting ...
(Boost.Redis) check-health-op: Operation canceled (async_ping), Operation canceled (async_check_timeout).
(Boost.Redis) runner-op: Got RESP3 simple-error. (async_run_all), Operation canceled (async_health_check) Success (async_hello).
(Boost.Redis) Connection lost: Got RESP3 simple-error.
```

However, if I remove this line  `req.push("QUIT");`, everything works as expected:
```
(Boost.Redis) run-all-op: resolve addresses 192.168.88.81:26379
(Boost.Redis) run-all-op: connected to endpoint 192.168.88.81:26379
(Boost.Redis) writer-op: 147 bytes written.
(Boost.Redis) reader-op: 172 bytes read.
(Boost.Redis) hello-op: Success
(Boost.Redis) ping_op: error/cancelled (2).
(Boost.Redis) check-timeout-op: error/canceled. Exiting ...
(Boost.Redis) check-health-op: Operation canceled (async_ping), Operation canceled (async_check_timeout).
(Boost.Redis) reader-op: Operation canceled
(Boost.Redis) reader-op: error. Exiting ...
(Boost.Redis) writer-op: canceled (3). Exiting ...
(Boost.Redis) run-op: Operation canceled (reader), Success (writer)
(Boost.Redis) runner-op: Operation canceled (async_run_all), Operation canceled (async_health_check) Success (async_hello).
(Boost.Redis) Connection lost: Operation canceled
```

### LINKED ISSUE #237: How to reconnect using Sentinel ?
I'm trying to reconnect when the Redis server shutsdown using Redis Sentinel. 
When boost::redis::not_connected occurs while using async_exec, I want to connect to a valid Redis server using Sentinel. 
However, when async_exec uses config::reconnect_wait_interval, it keeps trying to connect to the previously disconnected Redis server.

If I don't use config::reconnect_wait_interval, I can get a new IP from Sentinel after boost::redis::not_connected is returned by async_exec. 
But due to Redis failover time, I get the same IP/Port as before. 
If I use this to execute async_exec, "Connection lost: No connection could be made because the target machine actively refused it" occurs, and then async_exec hangs.

Is there a way to connect to a valid Redis server using Sentinel?

### LINKED ISSUE #268: Add a server role check
Add a server role check as described in [this](https://github.com/boostorg/redis/issues/237#issuecomment-2957077888) comment.

### LINKED ISSUE #269: Support short lived connections
To implement sentinel based failover we need better support for short lived connections. At the moment there is no way to launch `async_exec` and `async_run` in parallel and cancel when `async_exec` completes, see [this](https://github.com/boostorg/redis/blob/develop/example/cpp20_resolve_with_sentinel.cpp#L44-L46) comment. 

Also, short lived connections don't need health-check but disabling them [here](https://github.com/boostorg/redis/blob/develop/include/boost/redis/config.hpp#L72) does not prevent their operations from being started [here](https://github.com/boostorg/redis/blob/develop/include/boost/redis/connection.hpp#L372-L377) which should be avoided if possible to improve compilation times.

Moreover, for should lived connections it should be possible to send both the `HELLO` command and the actual request in a single request object, this is currently not possible because the resp3-handshake can't be deactivated [here](https://github.com/boostorg/redis/blob/develop/include/boost/redis/connection.hpp#L370).

We might be able to address these two points by adding an `async_run_exec` function that only launches the request and the read and write operations.

anarthal:
close #237 
close #269 
close #268
close #229

### DIFF
--- a/test/print_node.hpp
+++ b/test/print_node.hpp
@@ -1,0 +1,28 @@
+
+/* Copyright (c) 2018-2025 Marcelo Zimbres Silva (mzimbres@gmail.com)
+ *
+ * Distributed under the Boost Software License, Version 1.0. (See
+ * accompanying file LICENSE.txt)
+ */
+
+#ifndef BOOST_REDIS_TEST_PRINT_NODE_HPP
+#define BOOST_REDIS_TEST_PRINT_NODE_HPP
+
+#include <boost/redis/resp3/node.hpp>
+#include <boost/redis/resp3/type.hpp>
+
+#include <ostream>
+
+namespace boost::redis::resp3 {
+
+template <class String>
+std::ostream& operator<<(std::ostream& os, basic_node<String> const& nd)
+{
+   return os << "node{ .data_type=" << to_string(nd.data_type)
+             << ", .aggregate_size=" << nd.aggregate_size << ", .depth=" << nd.depth
+             << ", .value=" << nd.value << "}";
+}
+
+}  // namespace boost::redis::resp3
+
+#endif  // BOOST_REDIS_TEST_SANSIO_UTILS_HPP
--- a/test/test_conn_sentinel.cpp
+++ b/test/test_conn_sentinel.cpp
@@ -19,6 +19,7 @@
 #include <boost/core/lightweight_test.hpp>
 
 #include "common.hpp"
+#include "print_node.hpp"
 
 #include <string>
 
@@ -26,18 +27,6 @@ namespace net = boost::asio;
 using namespace boost::redis;
 using namespace std::chrono_literals;
 using boost::system::error_code;
-
-// TODO: duplicated
-namespace boost::redis::resp3 {
-
-std::ostream& operator<<(std::ostream& os, node const& nd)
-{
-   return os << "node{ .data_type=" << to_string(nd.data_type)
-             << ", .aggregate_size=" << nd.aggregate_size << ", .depth=" << nd.depth
-             << ", .value=" << nd.value << "}";
-}
-
-}  // namespace boost::redis::resp3
 
 namespace {
 
--- a/test/test_exec_one_fsm.cpp
+++ b/test/test_exec_one_fsm.cpp
@@ -18,6 +18,8 @@
 #include <boost/core/lightweight_test.hpp>
 #include <boost/system/error_code.hpp>
 
+#include "print_node.hpp"
+
 #include <iterator>
 #include <ostream>
 #include <string_view>
@@ -79,9 +81,7 @@ struct adapter_event {
          case parse_event::init: return os << "adapter_event{ .type=init }";
          case parse_event::done: return os << "adapter_event{ .type=done }";
          case parse_event::node:
-            return os << "adapter_event{ .type=node, .data_type=" << to_string(value.node.data_type)
-                      << ", .aggregate_size=" << value.node.aggregate_size
-                      << ", .depth=" << value.node.depth << ", .value=" << value.node.value << " }";
+            return os << "adapter_event{ .type=node, .node=" << value.node << " }";
          default: return os << "adapter_event{ .type=unknown }";
       }
    }
--- a/test/test_low_level_sync_sans_io.cpp
+++ b/test/test_low_level_sync_sans_io.cpp
@@ -13,6 +13,8 @@
 #include <boost/redis/resp3/type.hpp>
 #include <boost/redis/response.hpp>
 
+#include "print_node.hpp"
+
 #define BOOST_TEST_MODULE low_level_sync_sans_io
 #include <boost/test/included/unit_test.hpp>
 
@@ -338,24 +340,14 @@ BOOST_AUTO_TEST_CASE(check_counter_adapter)
 namespace boost::redis::resp3 {
 
 template <class String>
-std::ostream& operator<<(std::ostream& os, basic_node<String> const& nd)
-{
-   os << "type: " << to_string(nd.data_type) << "\n"
-      << "aggregate_size: " << nd.aggregate_size << "\n"
-      << "depth: " << nd.depth << "\n"
-      << "value: " << nd.value << "\n";
-   return os;
-}
-
-template <class String>
 std::ostream& operator<<(std::ostream& os, basic_tree<String> const& resp)
 {
-   for (auto const& e: resp)
+   for (auto const& e : resp)
       os << e << ",";
    return os;
 }
 
-}
+}  // namespace boost::redis::resp3
 
 node from_node_view(node_view const& v)
 {
@@ -370,7 +362,7 @@ node from_node_view(node_view const& v)
 tree from_flat(flat_tree const& resp)
 {
    tree ret;
-   for (auto const& e: resp.get_view())
+   for (auto const& e : resp.get_view())
       ret.push_back(from_node_view(e));
 
    return ret;
@@ -379,12 +371,11 @@ tree from_flat(flat_tree const& resp)
 tree from_flat(generic_flat_response const& resp)
 {
    tree ret;
-   for (auto const& e: resp.value().get_view())
+   for (auto const& e : resp.value().get_view())
       ret.push_back(from_node_view(e));
 
    return ret;
 }
-
 
 // Parses the same data into a tree and a
 // flat_tree, they should be equal to each other.
--- a/test/test_multiplexer.cpp
+++ b/test/test_multiplexer.cpp
@@ -15,6 +15,7 @@
 #include <boost/assert/source_location.hpp>
 #include <boost/core/lightweight_test.hpp>
 
+#include "print_node.hpp"
 #include "sansio_utils.hpp"
 
 #include <iostream>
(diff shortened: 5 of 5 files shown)


anarthal:
Still an early prototype, I want to check that my CI approach is viable.

include/boost/redis/config.hpp
```diff
@@ -24,21 +25,142 @@ struct address {
...
+ *
+ *  @param a Left hand side address.
+ *  @param b Right hand side address.
+ */
+inline bool operator==(address const& a, address const& b)
+{
+   return a.host == b.host && a.port == b.port;
+}
+
+/** @brief Compares two addresses for difference.
```
mzimbres:
```suggestion
/** @brief Compares two addresses for inequality.
```

include/boost/redis/detail/sentinel_resolve_fsm.hpp
mzimbres, line 1: Small thing that I think is more readable. Instead of

```cpp
enum class sentinel_action_type {...};

class sentinel_action {
   sentinel_action_type type_;
   sentinel_action(sentinel_action_type type) noexcept
}
```

I would write

```cpp

class sentinel_action {
   enum class type {...};
   type type_;
   sentinel_action(type type) noexcept
}
```

include/boost/redis/detail/sentinel_resolve_fsm.hpp
```diff
@@ -0,0 +1,92 @@
...
+   sentinel_action(const address& addr) noexcept
+   : type_(sentinel_action_type::connect)
+   , connect_(&addr)
+   { }
+
+   static sentinel_action request() { return {sentinel_action_type::request}; }
+
+   sentinel_action_type type() const { return type_; }
+
+   system::error_code error() const
```
mzimbres, line 51:
```suggestion
   [[nodiscard]]
   system::error_code error() const
```

include/boost/redis/detail/exec_one_fsm.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+   exec_one_action(system::error_code ec) noexcept
+   : type{exec_one_action_type::done}
+   , ec{ec}
+   { }
+};
+
+class exec_one_fsm {
+   int resume_point_{0};
+   any_adapter resp_;
```
mzimbres:
```suggestion
   any_adapter adapter_;
```

include/boost/redis/impl/run_fsm.ipp
```diff
@@ -41,45 +46,44 @@ inline void compose_ping_request(const config& cfg, request& to)
...
 }
 
+inline any_address_view get_server_address(const connection_state& st)
+{
+   if (!st.cfg.unix_socket.empty()) {
+      return any_address_view{st.cfg.unix_socket};
+   } else {
+      return {st.cfg.addr, st.cfg.use_ssl};
+   }
+}
```
mzimbres, line 70: For readability I would remove the `!` in `if (!st.cfg.unix_socket.empty())` and swap the return statements.

EDIT: Or perhaps add a `config` member `use_unix_sockets()`.

include/boost/redis/impl/sentinel_resolve_fsm.ipp
```diff
@@ -0,0 +1,184 @@
...
+
+         // When asking for replicas, we might get no replicas
+         if (st.cfg.sentinel.server_role == role::replica && resp.replicas.empty()) {
+            log_sentinel_error(st, idx_, "the configured master has no replicas");
+            continue;
+         }
+
+         // Store the resulting address in a well-known place
+         if (st.cfg.sentinel.server_role == role::master) {
+            st.cfg.addr = resp.master_addr;
```
mzimbres, line 126: IMO this assignment should be moved to the `run_op` right after `async_sentinel_resolve` resumes to make that function more readable. It would be more intuitive if `async_sentinel_resolve` complete with the value instead of changing the `connection_impl` state. Can we change the completion signature to `void(error_code, address)` to achieve that?

EDIT: We could also extend the completion signature with a third parameter which are the new sentinel addresses which is also use in run_op. This way the `async_sentinel_resolve` would not change any state in connection_impl.

include/boost/redis/impl/sentinel_utils.hpp
```diff
@@ -0,0 +1,277 @@
...
+   std::vector<address>& out)
+{
+   const auto* it = first;
+   ignore_unused(last);
+
+   // The root node must be an array
+   BOOST_ASSERT(it != last);
+   BOOST_ASSERT(it->depth == 0u);
+   if (it->data_type != resp3::type::array)
+      return {error::invalid_data_type};
```
mzimbres: This is not the correct error code. The data type is valid, just not expected. From the current error codes `expects_resp3_aggregate`, is the one that describes better the problem, but feel free to add `expects_resp3_array` as we already have `expects_resp3_map`.

include/boost/redis/impl/sentinel_utils.hpp
```diff
@@ -0,0 +1,277 @@
...
+   ++it;
+
+   // Each element in the array represents a server
+   out.resize(num_servers);
+   for (std::size_t i = 0u; i < num_servers; ++i) {
+      // A server is a map (resp3) or array (resp2, currently unsupported)
+      BOOST_ASSERT(it != last);
+      BOOST_ASSERT(it->depth == 1u);
+      if (it->data_type != resp3::type::map)
+         return {error::invalid_data_type};
```
mzimbres:
```suggestion
         return {error::expects_resp3_map};
```

include/boost/redis/impl/sentinel_utils.hpp
```diff
@@ -0,0 +1,277 @@
...
+   cfg.sentinel.setup.push("SENTINEL", "GET-MASTER-ADDR-BY-NAME", cfg.sentinel.master_name);
+   if (cfg.sentinel.server_role == role::replica)
+      cfg.sentinel.setup.push("SENTINEL", "REPLICAS", cfg.sentinel.master_name);
+   cfg.sentinel.setup.push("SENTINEL", "SENTINELS", cfg.sentinel.master_name);
+
+   // Note that we don't care about request flags because this is a one-time request
+}
+
+// Parses a list of replicas or sentinels
+inline system::error_code parse_server_list(
```
mzimbres, line 47: - Can we move this to .ipp?
- Can we take a `std::vector<node>` instead and work with `.at()` function instead of iterators. We don't need performance here so I would prefer to be on the safer side.

include/boost/redis/impl/sentinel_utils.hpp
mzimbres: More background: The `invalid_data_type` was meant to be used when the wire value of the data type was not valid, i.e. not one of `$`, `+`, `-`, `%` etc. Perhaps we should also have a `unexpected_data_type` and drop the individual `unexpected_resp3_array`, `unexpected_resp3_map` etc. that is not scalable since it requires too many error codes.

mzimbres (approved):

include/boost/redis/impl/sentinel_utils.hpp
anarthal: I've added `error::expects_resp3_array` and `error::expects_resp3_string`. If these ever come to pass, they yield better information than the single error code.

include/boost/redis/impl/sentinel_resolve_fsm.ipp
anarthal: Moving the assignment or the update_sentinel_list to run makes run (which is already big) have even more responsibilities. The tests are already gigantic, so I'm not very keen on this.

I don't think adding a 2nd (or 3rd) completion parameter to the run FSM makes a lot of good, because the parameter needs to be there for code paths, but has a valid value only in one of it. Unless you have a very clear opinion on this, I'd prefer for this to stay as is.

include/boost/redis/impl/sentinel_utils.hpp
anarthal: > Can we move this to .ipp?

I don't think it makes sense because this is an `impl/` file. It gets included by `.ipp` files only, and is never seen by the end-user code.

anarthal: > Can we take a std::vector<node> instead and work with .at() function instead of iterators. We don't need performance here so I would prefer to be on the safer side.

I'm gonna defer this to #355. `flat_tree` contains the number of messages, and this makes parsing much easier. None of these asserts can be triggered unless there's a bug in the parser, so it's not much of a problem. I will address it though, since you're right about the security/performance tradeoff here.

PR #347, opened 2025-11-03
Adds release notes for Boost 1.89 and Boost 1.90

anarthal:


anarthal:
Once we merge this, I'll submit a patch to the Boost website, which is where the official release notes should be. We haven't been doing this for past releases, never late to start.

PR #370, opened 2025-12-06
Cleans up test_low_level

anarthal:


anarthal:
This is still in-progress, but the idea is cleaning up test_low_level a little bit and increasing the coverage we have for parsers and adapters. My idea is splitting test_low_level into

* Parser tests. Given a valid input string, check that the correct sequence of output nodes is produced. No adapters involved.
* Parser error tests. Given an invalid input string, check that the correct error code is produced. No adapters involved.
* Single-field adapter tests. Given a valid sequence of nodes, check that the correct C++ value is generated.
* Single-field adapter error tests. Given a sequence of nodes that is valid RESP3 but is incompatible with the adapter (e.g. we expect an int and got an array), check that the adequate error code is obtained.

PR #375, opened 2025-12-19
Adds support for PubSub state restoration

### LINKED ISSUE #367: Pubsub state restoration
Currently, every time a reconnection happens, we cancel `async_receive2` so the user has a chance to re-issue all the SUBSCRIBE commands. This is not very user-friendly. Other clients do this automatically.

The easiest way would be adding the SUBSCRIBE commands to the setup request, but this wouldn't allow dynamically subscribing/unsubscribing to channels. So we need another mechanism to cover this.

anarthal:
Adds config::restore_pubsub_state and command_context
Deprecates the boost_redis_to_bulk extension point taking a std::string& argument, in favor of an extension point with the same name taking a command_context argument

close #367

### DIFF
--- a/test/test_run_fsm.cpp
+++ b/test/test_run_fsm.cpp
@@ -565,7 +565,7 @@ void test_setup_ping_requests()
    const std::string_view
       expected_setup = "*5\r\n$5\r\nHELLO\r\n$1\r\n3\r\n$4\r\nAUTH\r\n$3\r\nfoo\r\n$3\r\nbar\r\n";
    BOOST_TEST_EQ(fix.st.ping_req.payload(), expected_ping);
-   BOOST_TEST_EQ(fix.st.cfg.setup.payload(), expected_setup);
+   BOOST_TEST_EQ(fix.st.setup_req.payload(), expected_setup);
 
    // Reconnect
    act = fix.fsm.resume(fix.st, error::empty_field, cancellation_type_t::none);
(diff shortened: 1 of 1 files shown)


include/boost/redis/config.hpp
```diff
@@ -345,6 +345,9 @@ struct config {
     *        @ref sentinel_config::addresses is not empty.
     */
    sentinel_config sentinel{};
+
+   // (TODO: document properly) Do we want to re-subscribe to channels on reconnection?
+   bool restore_pubsub_state = false;
```
mzimbres: I think there should be no option to deactivate restoration in the connection. We can offer a `request` option e.g. `request::config::track_subscriptions`.

include/boost/redis/detail/pubsub_state.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace boost::redis {
+
+class request;
+
+namespace detail {
+
+enum class pubsub_change_type;
+
+class pubsub_state {
```
mzimbres: Perhpas name it `subscriptions` or `subscription _tracker`?

include/boost/redis/impl/pubsub_state.ipp
```diff
@@ -0,0 +1,48 @@
...
+      case pubsub_change_type::unsubscribe:  channels_.erase(std::move(owning_channel)); break;
+      case pubsub_change_type::psubscribe:   pchannels_.insert(std::move(owning_channel)); break;
+      case pubsub_change_type::punsubscribe: pchannels_.erase(std::move(owning_channel)); break;
+      default:                               BOOST_ASSERT(false);
+   }
+}
+
+void pubsub_state::commit_changes(const request& req)
+{
+   for (const auto& ch : detail::request_access::pubsub_changes(req))
```
mzimbres:
```suggestion
   for (const auto& ch : request_access::pubsub_changes(req))
```

Issue #331, opened 2025-10-13
Rework async_receive to consume all elements in the channel


mzimbres:
As I work on the `generic_flat_response` implementation I noticed that `async_receive` and its sync counterpart `receive` are too cumbersome and inefficient to work with.  At the moment the general for of the receive loop looks like this

```cpp
for (error_code ec;;) {
   conn->receive(ec);
   if (ec == error::sync_receive_push_failed) {
      ec = {};
      co_await conn->async_receive(asio::redirect_error(asio::use_awaitable, ec));
   }

   if (ec)
      break;  // Connection lost, break so we can reconnect to channels.

   ...

   consume_one(resp);
}
```

The `async_receive` function does not actually receive a response, it just removes its size from the channel. However it does not make sense to consume only one message at time since the response passed to `conn.set_receive_response()` might contain multiple pushes (i.e. the channel size). If we can make `async_receive` consume/accumulate all elements in the channel, the loop gets simpler and there is no need for the `consume_one` function that is inefficient since it rotates the response (any performance gain made by `generic_flat_response` will be eclipsed by these rotations). The general form of the loop consuming pushes would become

```cpp
for (error_code ec;;) {
   // Accumulates all sizes in the channel.
   co_await conn->async_receive2(asio::redirect_error(asio::use_awaitable, ec));
   if (ec)
      break;  // Connection lost, break so we can reconnect to channels.

   for (auto const& elem: resp)
      // Process the node.

   resp.clear();
}
```

Things to do

  - Add an `async_receive2` function that consumes all messages in the channel.
  - Deprecate `async_receive`, `receive` and `consume_one`.

anarthal:
> I believe however that there is a race on the async_receive completion value

With the approach you're suggesting, I think this is somehow possible, too. At least, with this kind of approach:

```cpp
for (error_code ec;;) {
   // Accumulates all sizes in the channel.
   co_await conn->async_receive2(asio::redirect_error(asio::use_awaitable, ec));
   if (ec)
      break;  // Connection lost, break so we can reconnect to channels.

   for (auto const& elem: resp)
      co_await websocket.async_write(...); // async operation

   resp.clear();
}
```

With this setup, we might get "bogus" completions of `async_receive2`:

* `async_receive2` completes because 5 pushes arrived from the server.
* Processing them involves async operations, so more pushes might be received from the server in the meanwhile. Let's say by the time we arrived to the last push, another push is received.
* The loop would continue, consuming that last push (* see note).
[...]

PR #380, opened 2026-01-14
Reworks test_conn_push and test_conn_push2

anarthal:
Adds a test to verify that async_receive (v1) is cancelled on reconnection
Adds a test to verify that a reconnection is triggered when the receive adapter generates an error
Makes the unsubscribe and push adapter tests live only in test_conn_push2, since they test functionality common to async_receive and async_receive2
Migrates all tests to lightweight_test

Entails no functional change

### DIFF
--- a/test/test_conn_push.cpp
+++ b/test/test_conn_push.cpp
@@ -20,7 +20,7 @@
 #include "common.hpp"
 
 #include <cstddef>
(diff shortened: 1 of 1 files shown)


anarthal:
I'd like to merge this before tacking #331, to verify I don't break anything.

PR #381, opened 2026-01-16
Makes async_receive2 not cancel on reconnection

### LINKED ISSUE #331: Rework async_receive to consume all elements in the channel

anarthal:
async_receive2 is now only cancelled after calling connection::cancel() or using per-operation cancellation
Adds a restriction to only have one outstanding async_receive2 operation per connection
Adds error::already_running
Adds support for asio::cancel_after for async_receive2
Deprecates cancel(operation::receive)
Adds more documentation to async_receive2

close #331

### DIFF
--- a/tools/ci.py
+++ b/tools/ci.py
@@ -134,7 +134,7 @@ def _build_b2_distro(
     _run([
         _b2_command,
         '--prefix={}'.format(_b2_distro),
-        '--with-system',
+        '--with-headers',
         'toolset={}'.format(toolset),
         '-d0',
         'install'


anarthal:
Still lacking unit tests for the FSM.

include/boost/redis/connection.hpp
```diff
@@ -785,27 +807,68 @@ class basic_connection {
...
+    * eventually timeout. To avoid that apps that expect server pushes
+    * should call this function continuously in a loop.
+    *
+    * This function should be used instead of the deprecated @ref async_receive.
+    * It differs from `async_receive` in the following:
+    *
+    * @li `async_receive` is designed to consume a single push message at a time.
+    *     This can be inefficient when receiving lots of server pushes.
+    *     `async_receive2` is batch-oriented, and will only suspend once if
+    *     several pushes are received in a single network packet.
```
mzimbres:
```suggestion
    *     256 push messages accumulate internally.
```

anarthal: Hmmm, no. The reader suspends if 256 push messages accumulate. `async_receive2` is the consumer side. I'll try to rephrase this though.

PR #385, opened 2026-01-21
Adds a way to iterate over full RESP3 messages

anarthal:


anarthal:
@mzimbres I'm trying to make working with RESP3 messages easier. What do you think of this? It's a forward range that allows you to consume a flat tree by breaking it into individual messages. It's much easier to use than iterating the raw range, and it should have the same performance.

I've written this in an attempt to make processing pushes easier, as an alternative to #366.

PR #390, opened 2026-01-27
Adds a convenient way of iterating PubSub messages

### LINKED ISSUE #349: The correct parameter type to `set_receive_response` is neither documented nor enforced.
currently `connection::set_receive_response` can't fail, and it will accept a reference to any `Response` object.

If that `Response` object can't accommodate what we get from the server, it cancels the channel and `async_receive` returns cancelled. This took me a couple hours to figure out in a coworker's code just now.

I'm having a hard time using either the Redis documentation or the Boost.Redis documentation figuring out if it is _ever_ ok to use a fixed response type here and if so what that fixed response type should be.

If a fixed response is not allowed, this should at least be documented in the `set_receive_response` page in the reference, and ideally the method should be made to only accept a reference to a generic response object.

If a fixed response _is_ allowed, the type(s) of fixed response(s) you can use should be documented somewhere and/or shown in example code. And, if possible, the method should be made to only accept valid ones (although this seems likely tricky).

anarthal:
Adds push_view and push_parser. This is an input view range that allows parsing PubSub messages from a node tree.
Adds a doc page on how to use server pushes.
Adds reference docs to set_receive_response.

close #349

### DIFF
--- a/doc/modules/ROOT/pages/reference.adoc
+++ b/doc/modules/ROOT/pages/reference.adoc
@@ -59,6 +59,10 @@ xref:reference:boost/redis/generic_flat_response.adoc[`generic_flat_response`]
 
 xref:reference:boost/redis/consume_one-08.adoc[`consume_one`]
 
+xref:reference:boost/redis/push_view.adoc[`push_view`]
+
+xref:reference:boost/redis/push_parser.adoc[`push_parser`]
+
 
 |
 xref:reference:boost/redis/adapter/boost_redis_adapt.adoc[`boost_redis_adapt`]


include/boost/redis/impl/push_parser.ipp
mzimbres, line 1: My only concern with this PR is that this push parser skips non-push messages silently. This might make it impossible to understand user issues. As we iterate over the messages, would it be possible to the iterator to push a reference to the skipped nodes into the push_parser. then we could at least log them after the loop if necessary

```cpp
      auto parser = push_parser(resp.value());
      for (push_view elem : parser) {
           ...
      }
      for (auto node : parser.get_skipped()) {
           ...
      }
```

We could provide a macro to simplify that logging.

PR #404, opened 2026-03-29
Prevents coroutine examples from hanging in C++17 builds

anarthal:
Fixes an ODR violation that caused C++20 examples to hang indefinitely in C++17 builds. This was caused by mixing code using `-std=c++17` and `-std=c++20` in different translation units.


anarthal:
This fixes a set of CI failures that have been happening for the last weeks.

PR #409, opened 2026-04-25
Removes the standalone CMake workflow

anarthal:
All integration testing now happens via the superproject, gated by BOOST_REDIS_INTEGRATION_TESTS
Added namespacing to CMake functions and targets to avoid collisions
Increases the maximum CMake supported version to 4.2


anarthal:
The aim of this PR is simplifying the CMake and fixing the build failures we're currently seeing in CI. If I'm not mistaken, these are happening because we're calling `add_subdirectory` on other Boost libraries directly, which we shouldn't be doing.

To effectively use the superproject workflow for development, you can use something akin to (suppose you're in `$BOOST_ROOT/libs/redis`):

```
mkdir __build
cd __build
cmake \
  -DBUILD_TESTING=ON \
  -DBOOST_INCLUDE_LIBRARIES=redis \
  -DBOOST_REDIS_INTEGRATION_TESTS=ON
  ../../..
make -j8 tests
```

Or any IDE-driven equivalent.

@mzimbres let me know if this would work for you.

PR #411, opened 2026-04-30
Guards benchmarks behind BOOST_REDIS_INTEGRATION_TESTS

anarthal:
Namespaces CMake target names in benchmarks
Guards benchmarks in the main CMake by BOOST_REDIS_INTEGRATION_TESTS. They are only built to prevent code rotting, and shouldn't be built by superproject builds except in our CIs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,12 +49,13 @@ if(BUILD_TESTING)
 
   # Tests and common utilities
   add_subdirectory(test)
-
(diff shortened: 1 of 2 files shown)


mzimbres:
I am running benchmarks in another project https://github.com/mzimbres/redis-cli-comp. It does not fit here because I also implement in other languages and that would make redis in boostorg bloated. Soon we will be able to delete the benchmarks directory. Thanks.

mzimbres (approved):

PR #415, opened 2026-05-02
Implements lazy buffer rotations

mzimbres:
The read buffer won't rotate data if it can be increased without allocating more memory. The table below compares the performance before and after the changes

Strategy| time(s) | %usr  | %sys | %CPU  | rotation
------|---------| ------|------|-------|------------
Eager | 92.22   | 67.37 | 2.92 | 70.29 |   6.7 Gbps
Lazy  | 91.16   | 60.15 | 2.82 | 62.97 |  99.0 kbps

The biggest gain in the benchmark above is in CPU usage because the overall time cannot be decreased much since the client is far from saturating the CPU.

### DIFF
--- a/include/boost/redis/config.hpp
+++ b/include/boost/redis/config.hpp
@@ -322,8 +322,8 @@ struct config {
 
    /** @brief Maximum size of the socket read-buffer in bytes.
     *  
-    * Sets a limit on how much data is allowed to be read into the
-    * read buffer. It can be used to prevent DDOS.
+    * Sets a limit on how large the read-buffer is allowed to grow. It can be
+    * used to prevent DDOS.
     *
     * When using Sentinel, this setting applies to masters, replicas and Sentinels.
     */
--- a/include/boost/redis/connection.hpp
+++ b/include/boost/redis/connection.hpp
@@ -301,7 +301,7 @@ struct exec_one_op {
    void operator()(Self& self, system::error_code ec = {}, std::size_t bytes_written = 0u)
    {
       exec_one_action act = fsm_.resume(
-         conn_->st_.mpx.get_read_buffer(),
+         conn_->st_.mpx,
          ec,
          bytes_written,
          self.get_cancellation_state().cancelled());
--- a/include/boost/redis/detail/exec_one_fsm.hpp
+++ b/include/boost/redis/detail/exec_one_fsm.hpp
@@ -21,7 +21,7 @@
 
 namespace boost::redis::detail {
 
-class read_buffer;
+class multiplexer;
 
 // What should we do next?
 enum class exec_one_action_type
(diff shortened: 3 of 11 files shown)


anarthal:
Thanks. Will have a look soon.

anarthal (commented):
.

anarthal:
Since GH seems buggy, I'll place my review in a comment to avoid losing it:

- https://github.com/boostorg/redis/pull/415/changes#diff-cf67e521e151f16e639d371d5b972fe0aa2eb6680a855608648b706adb456ed6R34: This is safer if written as `buffer_.size() > cfg_.max_size - append_size;` because `buffer_.size() + append_size` may technically overflow, leading to UB, while `cfg_.max_size >= append_size` should always hold (although I suggest adding a `BOOST_ASSERT`).
- https://github.com/boostorg/redis/pull/415/changes#diff-cf67e521e151f16e639d371d5b972fe0aa2eb6680a855608648b706adb456ed6R37: If `erase()` didn't really erase anything (because `consumed_ == 0`), rotated should be zero, rather than `buffer_.size`).
- https://github.com/boostorg/redis/pull/415/changes#diff-f3f96a8f99f99ba2e1ea212bd5387b46fd9e49460add831c3cbabf203f62b8fbR59: I don't think this is about right - the test should pass without reserving over the configured capacity (it defeats the purpose of a max capacity)
- https://github.com/boostorg/redis/pull/415/changes#diff-f3f96a8f99f99ba2e1ea212bd5387b46fd9e49460add831c3cbabf203f62b8fbR80: typo: consummed (should be consumed)
- https://github.com/boostorg/redis/pull/415/changes#diff-f3f96a8f99f99ba2e1ea212bd5387b46fd9e49460add831c3cbabf203f62b8fbR198: This should check `capacity()` to verify that we didn't realloc
- `multiplexer::prepare_read()` must not be `noexcept`, since it can trigger a realloc now.

anarthal:
I've confirmed with vtune that the reader now takes less proportion of CPU time. Memmove no longer appears in the flame graph. Feel free to merge this when you want.

Something that'd be interesting is rewriting the read buffer without a vector, directly with a `unique_ptr<char>`. My experience is that this improves performance more than expected. But we can defer this to later.

mzimbres:
Thanks Ruben, I addressed all comments. AppVeyor is reporting
```
Build started
git clone -q https://github.com/boostorg/redis.git C:\projects\redis-bwe2m
git fetch -q origin +refs/pull/415/merge:
git checkout -qf FETCH_HEAD
The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```
Do you know how to fix this?

anarthal:
Appveyor is still not functional, please ignore whatever it reports for now.

mzimbres:
> My experience is that this improves performance more than expected. But we can defer this to later.

I agree this should be done later.

PR #416, opened 2026-05-03
Replaces Boost.Test by boost::core::lightweight_test

anarthal:


anarthal:
All the tests that we've been adding since some time have used lightweight_test, instead of Boost.Test. Since we had a CI failure affecting the superproject because of Boost.Test, I thought I'd complete the migration completely.

mzimbres (approved):

PR #429, opened 2026-06-18
Updates doc build dependencies to avoid vulnerabilities

anarthal:
Updates antora to 3.1.15
Upgrades Axios from 1.13, which had a critical security vulnerability, to 1.16

### DIFF
--- a/doc/package-lock.json
+++ b/doc/package-lock.json
@@ -5,18 +5,18 @@
   "packages": {
     "": {
       "dependencies": {
+        "@cppalliance/antora-cpp-reference-extension": "^0.1.1",
         "@cppalliance/antora-downloads-extension": "^0.0.2",
-        "@cppalliance/antora-cpp-reference-extension": "^0.1.0",
-        "antora": "^3.1.10"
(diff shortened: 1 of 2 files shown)


mborland:
Since I saw you post about this on the ML, have you considered enabling Dependabot on the repo? Then these dependencies and vulnerabilities are automatically taken care of. It's worked pretty well for me on Decimal. You enable it in the settings and then you can put together a trivial yaml file like this: https://github.com/boostorg/decimal/blob/develop/.github/dependabot.yml. Grouped updates really reduces noise.

anarthal:
I see. Thanks for sharing @mborland. The neat thing is that I've been receiving emails from dependabot since forever, and I've always thought they related to the benchmark programs and ignoring them. Re-reading them again, it looks like the tokio version used in the Rust benchmarks is also flagged. I suspect if I enable dependabot now, it will try to fix it, too.

@mzimbres I recall that your intention was to completely extract the benchmarks out of this repo, which would solve the problem. How is it going? Can I help anyhow?

PR #434, opened 2026-07-25
Provide adapter to store RESP3 maps in vector<pair<T, U>>

mzimbres:


include/boost/redis/adapter/detail/adapters.hpp
```diff
@@ -585,7 +635,6 @@ class wrapper<result<T>> {
    : result_(t)
    {
       if (result_) {
-         result_->value() = T{};
```
mzimbres, line 588: I actually think I will undo this change as it will break user code. We should not erase data on behalf of the user and removing this line makes it possible to the user to append data to the container. The problem is that it will break user code and cause weird errors since not all users clear response objects before passing them to `asyc_exec`. It is something hard to change now.

anarthal: I agree - I think we've settled on not clearing data for the other adapters, and clearing for this one seems confusing.

include/boost/redis/adapter/detail/adapters.hpp
```diff
@@ -421,6 +426,46 @@ class vector_impl {
...
+public:
+   void on_value_available(Result&) { }
+   void on_init() { }
+   void on_done() { }
+
+   template <class String>
+   void on_node(Result& res, resp3::basic_node<String> const& nd, system::error_code& ec)
+   {
+      if (is_aggregate(nd.data_type)) {
+         auto const m = element_multiplicity(nd.data_type);
```
anarthal, line 449: Shouldn't we require a multiplicity of exactly 2? What happens if the server sends us an array of 5 elements, for instance?

include/boost/redis/adapter/detail/adapters.hpp
anarthal, line 138: I think this is technically missing a `else { op.reset(); }`

anarthal:
This covers `vector<pair<T, U>>` nicely, but doesn't address the new C++23 associative containers, abseil or Boost.Unordered. After this gets merged, do you want me to try to come up with an approach that covers these, or are we not aiming to cover them?

include/boost/redis/adapter/detail/adapters.hpp
mzimbres: I thought I wouldn't be so strict. By not requiring `2` the user would be able to read a RESP3 array into a `vector<pair<U,V>>`, which for odd sizes would let the last element undefined as you noticed. Perhaps you are correct and we should exit with `expects_resp3_map`, which might catch some usage errors earlier.

mzimbres:
> This covers `vector<pair<T, U>>` nicely, but doesn't address the new C++23 associative containers, abseil or Boost.Unordered. After this gets merged, do you want me to try to come up with an approach that covers these, or are we not aiming to cover them?

Sure, thanks. It might not be trivial as I think there isn't a single concept that covers all e.g. map containers optimally.

